Issue #1, opened 2017-10-26
Is someone allowed to create a project when not logged in?


jingjingwang:
Right now "projects/signup" is accessible without being logged in. (We will get an error if click on "submit" though)

marlonkeating:
Yes, even though the 'ADD A PROJECT' link isn't visible when users aren't logged in, the url is still accessible.  Ideally, we'll want to redirect to the login page, which will forward to the Add Project page.

PR #2, opened 2017-10-28
Add user profile

jingjingwang:


jingjingwang:
Also solves #1.

civictechprojects/templates/project_signup.html
```diff
@@ -69,32 +69,32 @@ <h2 class="form-group subheader">DETAILS</h2>
 
             <div class="form-group">
                 <label for="brief_description">Describe This Project</label>
-                <textarea class="form-control" id="brief_description" placeholder="This will appear as project introduction" rows=3></textarea>
+                <textarea class="form-control" name="brief_description" placeholder="This will appear as project introduction" rows=3></textarea>
```
marlonkeating: This wasn't added by your change, but I noticed it during testing.  Name should be "project_description" in order for it to be saved by the view.

jingjingwang:
Added a few fixes. One thing that I'm not entirely sure is that, should we use an email address as a username. Few places would need to be revisited if one day we changed the design.

marlonkeating:
User profile design has changed, so closing this review.

PR #5, opened 2017-11-12
adding linux instructions

moonpatroller:


README.md
```diff
@@ -1,3 +1,32 @@
...
+
+# checkout the repo
+git clone https://github.com/DemocracyLab/CivicTechExchange.git
+
+# create and activate venv
+cd CivicTechExchange/
+python -m venv myenv
+source myvenv/bin/activate
+
+# install python libraries
```
marlonkeating: Requirements.txt is already a single source of truth for python dependencies, so we should just say:
pip install -r requirements.txt

README.md
```diff
@@ -1,3 +1,32 @@
...
+pip install django
+pip install dj_database_url
+pip install djangorestframework
+pip install taggit
+pip install django-taggit
+pip install whitenoise
+pip install psycopg2
+pip install boto3
+
+# create database via django and start test server
```
marlonkeating: Need to include node/yarn steps as well.  

Installing Node.js:
curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
sudo apt-get install nodejs

Installing Yarn package manager:
curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | sudo apt-key add -
echo "deb https://dl.yarnpkg.com/debian/ stable main" | sudo tee /etc/apt/sources.list.d/yarn.list
sudo apt-get update && sudo apt-get install yarn

Building/Collecting static assets before runserver:
npm run build

README.md
moonpatroller: It should have that now.

PR #14, opened 2017-12-05
Improvements to Filtering projects by Tag

lowellbander:
Plenty of work remains, but these changes accomplish the following:

* client-side Issue Areas have been refactored to just be a subset of Tags, and are fetched from the server-side definition
* search results are unioned rather than intersected
* display applied tag filters (see video below)

demo: https://imgur.com/a/Zm9VX


common/components/stores/ProjectSearchStore.js
```diff
@@ -80,17 +83,13 @@ class ProjectSearchStore extends ReduceStore<State> {
...
+  _loadProjects(state: State): State {
     const url = [
       '/api/projects?',
-      keyword ? '&keyword=' + keyword : null,
-      issueArea
-        ? '&issueArea=' + lodash.snakeCase(issueArea)
+      state.keyword ? '&keyword=' + this.getState().keyword : null,
+      state.tags
+        ? '&tags='
+          + state.tags.map(tag => tag.tagName).join(',').replace('-', '_')
```
marlonkeating: Why replace the tag name dashes with underscores?  We use dashes in tag names on the backend.

common/components/stores/ProjectSearchStore.js
```diff
@@ -80,17 +83,13 @@ class ProjectSearchStore extends ReduceStore<State> {
...
   }
 
-  _clearProjects(state: State): State {
-    return state.set('projects', null);
-  }
-
-  _loadProjects(state: State): void {
-    const {keyword, issueArea} = state;
+  _loadProjects(state: State): State {
     const url = [
```
marlonkeating: Props for being able to cram all the url building logic into a single expression, but I'm wondering if the more boring pattern:
if(state.x) {
  url += '&x=' + yaddayadda;}
might be more readable and maintainable in the long run.

common/components/stores/ProjectSearchStore.js
lowellbander: Previous to 55d2c1a29ce68c02a883d4303b1878d1b2cea4f7, tags used underscores. Will update UI to reflect backend change.

lowellbander: resolved by 878e7756fca87431eadc179a31342c85b4faa89f

common/components/stores/ProjectSearchStore.js
lowellbander: ty for the props <3 :3

I've refactored some in d60801a, lemme know whatcha think

civictechprojects/views.py
```diff
@@ -143,21 +125,13 @@ def projects_list(request):
...
-                )
-            )
-        if 'issueArea' in query_params:
-            projects = (
-                projects
-                .filter(
-                    project_issue_area__name__in=query_params['issueArea'],
-                )
-            )
+        projects = (
```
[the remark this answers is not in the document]
lowellbander: I see what you're saying about complicated query logic. I think this is partly because Python handles ternaries as

`resultIfTrue if cond else resultIfFalse`

rather than

`cond ? resultIfTrue : resultIfFalse`

which I agree is a little weird to grok.

However, I don't think the query logic can be refactored as suggested. The current form of the query construction is

`filtered_results if filters_exist else all_results`

so as to deal with the case of an empty result set. If filters were passed to the API and the filtered result set is empty, then we should return an empty result set. However, if the result set is empty because no filters were passed, then we should return a result containing all possible projects.

Lemme know if I'm not making sense.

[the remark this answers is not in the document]
marlonkeating: Ah, right, we need to chain filters in this case so you could do something like:
```
projects = Project.objects
if x:
  projects = projects.filter_by_x(x)
if y:
  projects = projects.filter_by_y(x)
```

common/components/stores/ProjectSearchStore.js
marlonkeating: Looks good!

PR #15, opened 2017-12-09
Link project cards to project info pages

lowellbander:


common/components/componentsBySection/FindProjects/ProjectCard.jsx
```diff
@@ -31,19 +31,28 @@ class ProjectCard extends React.PureComponent<Props> {
     );
   }
 
+  _goToDetailsView(): void {
+    window.location.replace('/project/' + this.props.project.id);
```
marlonkeating: Can we inject a regular link onto the cards?  Usability-wise, people will want to do things like ctrl-click, and see the hand when they hover over something that's a link.

lowellbander: Makes sense! will fix

PR #16, opened 2017-12-09
Add analytics to DLab

lowellbander:


civictechprojects/templates/new_index.html
```diff
@@ -19,6 +19,24 @@
     <link rel="stylesheet" type="text/css" href="/static/css/SubHeader.css"/>
   </head>
   <body>
+    <script>
+      window.fbAsyncInit = function() {
+        FB.init({
+          appId      : '137258853647008',
```
marlonkeating: Should this appId be factored out into an environment variable, so traffic from people's development environments don't pollute the results?

lowellbander: Totes see your concern -- we should be okay because we can create a segment (filter) when looking at data to filter our localhost, e.g.: 

image

marlonkeating: Okay, that should be good enough for the present.

PR #22, opened 2018-01-21
Edit Project page

marlonkeating:
Plus a number of fixes for files, and hooking up project thumbnails to the Find Projects page.


civictechprojects/forms.py
```diff
@@ -56,3 +58,44 @@ def create_project(request):
             thumbnail_json = json.loads(project_thumbnail_location)
             thumbnail = ProjectFile.from_json(project, FileCategory.THUMBNAIL, thumbnail_json)
             thumbnail.save()
+
+    @staticmethod
+    def edit_project(request, project_id):
+        project = Project.objects.get(id=project_id)
+
+        if not request.user.username == project.project_creator.username:
```
lowellbander: This seems like a potential security hole, because a bad actor can spoof the username in the request. Should you instead use some django built-ins to check if tokens are valid / if django claims the user is authorized to perform this action?

civictechprojects/forms.py
```diff
@@ -56,3 +58,44 @@ def create_project(request):
...
+            raise PermissionDenied()
+
+        form = ProjectCreationForm(request.POST)
+        project.project_description = form.data.get('project_description')
+        project.project_location=form.data.get('project_location')
```
lowellbander: nit: consistent formatting (spaces around assignment operator)

civictechprojects/forms.py
```diff
@@ -56,3 +58,44 @@ def create_project(request):
...
+
+        files_json_text = form.data.get('project_files')
+        if len(files_json_text) > 0:
+            files_json = json.loads(files_json_text)
+            ProjectFile.merge_changes(project, files_json)
+
+        project_thumbnail_location = form.data.get('project_thumbnail_location')
+        if len(project_thumbnail_location) > 0:
+            thumbnail_json = json.loads(project_thumbnail_location)
+            existing_thumbnail = ProjectFile.objects\
```
lowellbander: nit: I think you can avoid the backslash:

```
  existing_thumbnail = ProjectFile.objects.filter(
    file_project=project.id,
    file_category=FileCategory.THUMBNAIL.value,
  ).first()
```

civictechprojects/models.py
```diff
@@ -39,6 +39,39 @@ class Project(models.Model):
...
+
+    # Remove any tags that aren't part of the canonical list
+    @staticmethod
+    def remove_tags_not_in_list():
+        for project in Project.objects.all():
+            for issue in project.project_issue_area.all():
+                issue_tag = Tag.get_by_name(issue)
+                if not issue_tag:
+                    print('Removing invalid tag', issue, 'from project:', project.project_name)
+                    project.project_issue_area.remove(issue)
```
lowellbander: is it necessary to save this mutation?

civictechprojects/urls.py
```diff
@@ -19,7 +19,8 @@
 from . import views
 
 urlpatterns = [
-    url(r'^projects/signup/$', views.project_signup, name='project_signup'),
+    url(r'^projects/signup/(?P<project_id>[0-9]+)/$', views.project_edit, name='project_edit'),
```
lowellbander: if this is the URL for editing an existing project, seems weird that it has `signup` in the URL, no?

civictechprojects/views.py
```diff
@@ -42,25 +50,26 @@ def to_tag_map(tags):
...
+        return redirect('/signup')
+
+    try:
+        ProjectCreationForm.edit_project(request, project_id)
+    except PermissionDenied:
+        return HttpResponseForbidden()
+    return redirect('/index/?section=AboutProject&id=' + project_id)
 
 
+# TODO: Remove when React implementation complete
```
lowellbander: create a tracking task for this and reference it here?

common/components/chrome/MainHeader.jsx
```diff
@@ -31,6 +31,7 @@ class MainHeader extends React.PureComponent<{||}> {
     NavigationDispatcher.dispatch({
       type: 'SET_SECTION',
       section: Section.Landing,
+      url: '/index/?section=Landing'
```
lowellbander: why are you passing in this parameter here? Doesn't the `NavigationStore` handle this by itself: 

https://github.com/DemocracyLab/CivicTechExchange/blob/master/common/components/stores/NavigationStore.js#L35

common/components/chrome/SectionLink.jsx
```diff
@@ -42,6 +42,7 @@ class SectionLink extends React.PureComponent<Props> {
     NavigationDispatcher.dispatch({
       type: 'SET_SECTION',
       section: this.props.section,
+      url: '/index/?section=' + this.props.section
```
lowellbander: ditto

common/components/chrome/SubHeader.jsx
```diff
@@ -42,7 +42,7 @@ class SubHeader extends React.Component<{||}, State> {
         <span className={this._cx.get('rightContent')}>
           <a
             className={this._cx.get('createProject')}
-            href="/projects/signup"
+            href="/index/?section=CreateProject"
```
lowellbander: why href and not NavigationDispatcher?

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -0,0 +1,188 @@
...
+import type {FileInfo} from '../../common/FileInfo.jsx'
+import ImageUploadFormElement from '../../../components/forms/ImageUploadFormElement.jsx'
+import LinkList from '../../../components/forms/LinkList.jsx'
+import FileUploadList from '../../../components/forms/FileUploadList.jsx'
+import TagCategory from '../../common/tags/TagCategory.jsx'
+import TagSelect from '../../common/tags/TagSelect.jsx'
+import DjangoCSRFToken from 'django-react-csrftoken'
+import ProjectAPIUtils from '../../../components/utils/ProjectAPIUtils.js';
+import type {ProjectDetailsAPIData} from '../../../components/utils/ProjectAPIUtils.js';
+import _ from 'lodash'
```
lowellbander: i'd recommend importing this as `lodash` because it's a JS idiom to reserve `_` for denoting unused variables, like

```
try {
  foo()
} catch (_) {
  console.error('something bad happened')
}
```

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -0,0 +1,188 @@
...
+
+type FormFields = {|
+  project_name: ?string,
+  project_location: ?string,
+  project_url: ?string,
+  project_description: ?string,
+  project_links: Array<LinkInfo>,
```
lowellbander: why not `$ReadOnlyArray`?

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -0,0 +1,188 @@
...
+  project_files: Array<FileInfo>,
+  project_thumbnail: FileInfo
+|};
+
+type Props = {|
+  projectId: number
```
lowellbander: it might be better to handle these as `string`s, because if we deviate from allocating IDs sequentially and just choose them randomly from our possible ID-space, the numbers might be quite large (depending on how django is configured) in which case JS will round the numbers and our IDs will be not useful

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -0,0 +1,188 @@
...
+import FileUploadList from '../../../components/forms/FileUploadList.jsx'
+import TagCategory from '../../common/tags/TagCategory.jsx'
+import TagSelect from '../../common/tags/TagSelect.jsx'
+import DjangoCSRFToken from 'django-react-csrftoken'
+import ProjectAPIUtils from '../../../components/utils/ProjectAPIUtils.js';
+import type {ProjectDetailsAPIData} from '../../../components/utils/ProjectAPIUtils.js';
+import _ from 'lodash'
+
+type FormFields = {|
+  project_name: ?string,
```
lowellbander: if these are always at least empty strings, I think their type can be `string` rather than `?string`

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -0,0 +1,188 @@
...
+    this.checkFormValidity();
+  }
+  
+  onComponentChange(formFieldName: string, newValue: string): void {
+    this.state.formFields[formFieldName] = newValue;
+    this.checkFormValidity();
+  }
+  
+  checkFormValidity(): void {
+    var formFields = this.state.formFields;
```
lowellbander: nit: `const`

also you can take advantage of some fanciness and rewrite this as

`const {formFields} = this.state;`

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -0,0 +1,188 @@
...
+    
+    var valid = !_.isEmpty(formFields["project_name"]) && !_.isEmpty(formFields["project_description"]);
```
lowellbander: nit: const, and again fanciness to do:
```

const formIsVald = ...
this.setState({formIsValid});
```

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -0,0 +1,188 @@
...
+    this.setState({
+      formIsValid: valid
+    });
+  }
+  
+  render(): React$Node {
+    if(this.state.error) {
```
lowellbander: nit: can be terser with ternaries
`return this.state.error ? this._renderError() : this._renderForm();`

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -0,0 +1,188 @@
...
+    );
+  }
+  
+  _renderForm(): React$Node {
+    return (
+      <div className="EditProjectForm-root">
+        
+        <DjangoCSRFToken/>
+        
+        <div className="form-group">
```
lowellbander: is this `form-group` className being used anywhere?

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -0,0 +1,188 @@
...
+        </div>
+        
+        <div className="form-group">
+          <label htmlFor="project_location">Project Location</label>
+          <input type="text" className="form-control" id="project_location" name="project_location" maxLength="200"
+                 value={this.state.formFields.project_location} onChange={this.onFormFieldChange.bind(this, "project_location")}/>
+        </div>
+        <div className="form-group">
+          <label htmlFor="project_url">Website URL</label>
+          <input type="text" className="form-control" id="project_url" name="project_url" maxLength="200"
```
lowellbander: what's this className do? Also why the low maxLength value?

common/components/common/tags/TagSelect.jsx
```diff
@@ -0,0 +1,71 @@
...
+|};
+type State = {|
+  tags: Array<TagDefinition>,
+  selected: TagDefinition
+|};
+
+/**
+ * Dropdown selector for tags
+ */
+class TagSelect extends React.PureComponent<Props, State> {
```
lowellbander: nit: this name feels a little funky, how about `TagDropdown` or at least `TagSelector`

common/components/common/tags/TagSelect.jsx
```diff
@@ -0,0 +1,71 @@
...
+          value: tag.tag_name,
+          label: tag.display_name
+        }))
+      )
+      .then(options => this.setState({tags: options}));
+  }
+  
+  componentWillReceiveProps(nextProps: Props): void {
+    if(nextProps.value) {
+      this.setState({selected: {value:nextProps.value, label:_.find(this.state.tags, tag => tag.value === nextProps.value).label} || ""});
```
lowellbander: I think this is builtin and you don't need lodash here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -0,0 +1,188 @@
...
+    });
+  }
+  
+  onFormFieldChange(formFieldName: string, event: SyntheticInputEvent<HTMLInputElement>): void {
+    this.state.formFields[formFieldName] = event.target.value;
+    this.forceUpdate();
+    this.checkFormValidity();
+  }
+  
+  onComponentChange(formFieldName: string, newValue: string): void {
```
lowellbander: nit: function could be called something like `onFormFieldChange`

common/components/componentsBySection/MyProjects/MyProjectCard.jsx
```diff
@@ -38,6 +38,11 @@ class MyProjectCard extends React.PureComponent<Props> {
                 </tr>
                 <tr>In Progress</tr>
               </td>
+              <td className="MyProjectCard-column">
+                <tr className="MyProjectCard-header">
+                  <a href={'?section=EditProject&id=' + this.props.project.id}>Edit</a>
```
lowellbander: might be good to create some URLBuilder class that gets passed in a Section so that all these string literals don't break if we change our URL scheme?

common/components/controllers/EditProjectController.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Encapsulates form for editing projects
+ */
+class EditProjectController extends React.PureComponent<{||},State> {
+  constructor(props: Props): void {
+    super(props);
+  
+    this.state = {
+      projectId:  (new RegExp("id=([^&]+)")).exec(document.location.search)[1]
```
lowellbander: this seems kinda funky, like it should be gotten from a Prop or a Store or at least a Util devoted to URL inspection

civictechprojects/forms.py
marlonkeating: My understanding is that request.user is populated from the session cookie.  And indeed, looking at the request headers I can't see anything that contains user information in plain text.

civictechprojects/models.py
marlonkeating: No.  Adding and removing tags doesn't seem to require saving the base Project.

common/components/chrome/MainHeader.jsx
marlonkeating: We need to pass in parameters for viewing/editing projects, and the navigation system removed any parameters after the ?section= part.  This was the quickest way I saw to add in the url info those pages needed.  

If there's a scheme for doing this that fits better with the Flux paradigm, I'm definitely open to using it.

common/components/common/projects/EditProjectForm.jsx
marlonkeating: If we do change to an id scheme that would run afoul of JS rounding, I imagine we'd probably have to come back and edit this part anyways.

civictechprojects/views.py
marlonkeating: https://trello.com/c/gyaguFWh

common/components/common/projects/EditProjectForm.jsx
marlonkeating: It's part of bootstrap's css

common/components/common/projects/EditProjectForm.jsx
marlonkeating: form-control is another bootstrap styling.

And yeah, that is a low maxLength value for a url field.  Will change it to be 2000 in line with our other urls.

common/components/common/projects/EditProjectForm.jsx
marlonkeating: '_.' is also the idiomatic way to use the lodash/underscore libraries, and helpful for brevity.

common/components/common/projects/EditProjectForm.jsx
marlonkeating: Because the arrays are mutated on this form.

common/components/common/projects/EditProjectForm.jsx
marlonkeating: onFormFieldChange is already in use for traditional events which use SyntheticInputEvents, as opposed to Components which return string values.

common/components/common/projects/EditProjectForm.jsx
marlonkeating: I've seen you use that ES6 object destructuring syntax but keep forgetting to use it in my own work.  Will try it out!

civictechprojects/forms.py
marlonkeating: Done

civictechprojects/forms.py
marlonkeating: Done

civictechprojects/urls.py
marlonkeating: Changed

common/components/chrome/SubHeader.jsx
marlonkeating: Changed

common/components/common/projects/EditProjectForm.jsx
marlonkeating: Done

common/components/common/projects/EditProjectForm.jsx
marlonkeating: Done

common/components/common/tags/TagSelect.jsx
marlonkeating: Done

common/components/common/tags/TagSelect.jsx
marlonkeating: Done

common/components/componentsBySection/MyProjects/MyProjectCard.jsx
marlonkeating: Done

common/components/controllers/EditProjectController.jsx
marlonkeating: Moved to url library

PR #25, opened 2018-03-11
removed link underlining

alagrin92:
Test of update to files


marlonkeating:
Looks reasonable at first glance, but I'm still seeing link underlining here:
https://democracy-lab-staging-pr-25.herokuapp.com/index/?section=FindProjects

common/components/componentsBySection/FindProjects/ProjectCard.jsx
```diff
@@ -21,7 +25,7 @@ class ProjectCard extends React.PureComponent<Props> {
 
   render(): React$Node {
     return (
-      <a
+      <a style={styles}
```
marlonkeating: Is there a need to inject this style into the markup?

package.json
```diff
@@ -23,15 +21,17 @@
     "react": "^16.0.0",
     "react-bootstrap": "^0.31.5",
     "react-dom": "^16.0.0",
-    "react-select": "^1.2.0",
+    "react-select": "^1.2.1",
```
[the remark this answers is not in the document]
alagrin92: this one fixed the errors I was getting upon trying to run the build

common/components/componentsBySection/FindProjects/ProjectCard.jsx
alagrin92: This type of fix was the only one that actually was making sure all links at the same time were not underlining, unless there is a better method. Tried all I knew!

marlonkeating: Alright.  In that case can we remove the redundant lines in the .css?

alagrin92: Will do!

PR #27, opened 2018-04-11
Add header to login and signup

olgarose:
Made a new file 'no_link_base.html', which is exactly like 'base.html', but without the links for 'login' or 'signup' in the header. 

This needs to be refactored to have 'base.html' have no links (to be a true base) and a new html that extends base.html and has the 'login' and 'signup' links (currently being used on the 'forgot password' page).

Have the 'login' and 'signup' pages extending 'no_link_base.html', so they now have headers.


marlonkeating:
Branch has been abandoned, so closing this review.

PR #29, opened 2018-04-16
Reactify signup signin

olgarose:
Hook up signup and login with react


olgarose:
The links are now directing to the React controllers, but I need help with UI (specifically spacing between links)

PR #40, opened 2018-05-17
Contact Project button redirects the user to the sign in page.

StanislavLev:


common/components/common/projects/ContactProjectButton.jsx
```diff
@@ -83,27 +83,46 @@ class ContactProjectButton extends React.PureComponent<Props, State> {
...
+          </div>
+        );
+      } else {
+        return (
+          <div>
+            {this.state.buttonVisible
+              ? <Button
+                className="ProjectSearchBar-submit"
+                type="button"
+                disabled={!this.state.buttonDisabled}
```
marlonkeating: Make sure to set disabled to false above.

PR #41, opened 2018-05-22
add description to discover page

StanislavLev:
If the description will be short like a header, we can change <p> tag to <h*> and remove added css.


civictechprojects/static/css/DiscoverPageDescription.css
```diff
@@ -0,0 +1,4 @@
+.DiscoverPageDescription-root {
```
marlonkeating: Is this used?

StanislavLev: No, I think we'll need it only if we'll have long description with more than one paragraph.

PR #45, opened 2018-05-24
add edit button when owner look at his own project

StanislavLev:


common/components/common/projects/ContactProjectButton.jsx
```diff
@@ -82,44 +75,57 @@ class ContactProjectButton extends React.PureComponent<Props, State> {
...
 
   render(): ?React$Node {
+    const id = {'id':this.props.project.project_id};
     if(this.state) {
       if(CurrentUser.isLoggedIn()) {
-        return (
-          <div>
-            {this.state.buttonVisible
-              ? <Button
+        if(CurrentUser.userID() === this.props.project.project_creator){
```
marlonkeating: WIth the growing function size and multiple conditional expressions going on, this code is getting harder to follow.  At this point I would recommend breaking out each Button element into a separate render function.  For example:

if(CurrentUser.userID() === this.props.project.project_creator){
  return  _renderEditProjectButton();
}

If you want to see an example of where we do this, check out ProjectCard.jsx

StanislavLev:
I wasn't sure if it is good to render ContactProjectModal as a function or not, meanwhile it remains like it was.

PR #46, opened 2018-06-13
Added padding to the bottom of the page

Sadia254:


civictechprojects/static/css/projects.css
```diff
@@ -247,5 +247,5 @@ th, td {
 .Text-section {
   background-color: white;
   border: solid 1px #cecece;
-  padding: 10px;
+  padding: 50px;
```
marlonkeating: This is actually padding the inside of the text box, not the bottom.

civictechprojects/static/css/projects.css
```diff
@@ -248,4 +248,5 @@ th, td {
   background-color: white;
   border: solid 1px #cecece;
   padding: 10px;
+  margin-bottom: 50px;
```
marlonkeating: There's a bit too much space between items right now, I would go with a smaller number.

PR #48, opened 2018-06-20
Add url filter parameters to Find Project page

marlonkeating:
We want to be able to save and load project filters in the FindProjects url.  This change updates the url when filters are selected, and will apply those filters when the url is navigated to.


marlonkeating:
The filter behavior you describe is expected.  Basically, filters in the same category are applied on an OR basis.

PR #49, opened 2018-06-25
Location drop down

StanislavLev:


common/components/common/projects/EditProjectForm.jsx
```diff
@@ -160,16 +160,22 @@ class EditProjectForm extends React.PureComponent<Props,State> {
...
         <div className="form-group">
           <label htmlFor="project_url">Website URL</label>
           <input type="text" className="form-control" id="project_url" name="project_url" maxLength="2075"
                  value={this.state.formFields.project_url} onChange={this.onFormFieldChange.bind(this, "project_url")}/>
         </div>
+
+        <div className="form-group">
+          <label htmlFor="project_location">Project Location</label>
+          <select name="project_location" id="project_location" className="form-control" value={this.state.formFields.project_location} onChange={this.onFormFieldChange.bind(this, "project_location")}>
+            <option value={this.state.formFields.project_location}>{this.state.formFields.project_location}</option>
```
marlonkeating: Let's put an 'Other' option too.

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -174,6 +174,7 @@ class EditProjectForm extends React.PureComponent<Props,State> {
             <option value="Kirkland, WA">Kirkland, WA</option>
             <option value="Bellevue, WA">Bellevue, WA</option>
             <option value="Seattle, WA">Seattle, WA</option>
+            <option value="Other">Other</option>
```
marlonkeating: We don't really want to show 'Other' as the location on the About Project page, so could you add that bit to AboutProjectController.jsx to hide the text if it's Other?

PR #51, opened 2018-06-29
Left Side Filters

PeterBreen:
This PR:
- Moves filters to left, expandable column instead of dropdown (style still unfinished)
- Changes filters to be checkbox items
- Checkboxes can both add/remove filters
- Checkboxes respect saved queries (should have the appropriate boxes checked if you load a saved set of filters via url)
- shouldn't break dropdown functionality if it's used elsewhere


civictechprojects/static/css/ContextualCollapsible.css
```diff
@@ -0,0 +1,31 @@
+.ContextualCollapsible-root {
+  background: none;
+  border: none;
+  margin-top: 10px;
+  padding-bottom: 15px;
+  padding-top: 15px;
```
marlonkeating: IMO the UI looks better without any padding on top of the collapsible area.

PR #52, opened 2018-07-09
Upadte project add edit interface

StanislavLev:
"Upload project image" and  "Project Files" use the same
<FileUploadButton> so updated "Upload project image" as well.

### DIFF
--- a/common/components/common/projects/EditProjectForm.jsx
+++ b/common/components/common/projects/EditProjectForm.jsx
@@ -215,14 +215,17 @@ class EditProjectForm extends React.PureComponent<Props,State> {
                     value={this.state.formFields.project_description} onChange={this.onFormFieldChange.bind(this, "project_description")}></textarea>
         </div>
   
-        <PositionList elementid="project_positions" positions={this.state.formFields.project_positions}/>
-        
-        <LinkList elementid="project_links" links={this.state.formFields.project_links}/>
+        <div className="form-group">
+          <PositionList elementid="project_positions" positions={this.state.formFields.project_positions}/>
+        </div>
 
         <div className="form-group">
-          <label>Files</label>
+          <LinkList elementid="project_links" links={this.state.formFields.project_links}/>
+        </div>
+
+        <div className="form-group">
           <FileUploadList elementid="project_files" files={this.state.formFields.project_files}/>
-        </div>       
+        </div>
 
         <FormValidation
           validations={this.state.validations}
--- a/common/components/common/projects/EditProjectForm.jsx
+++ b/common/components/common/projects/EditProjectForm.jsx
@@ -217,10 +217,7 @@ class EditProjectForm extends React.PureComponent<Props,State> {
   
         <PositionList elementid="project_positions" positions={this.state.formFields.project_positions}/>
         
-        <div className="form-group">
-          <label>Links</label>
-          <LinkList elementid="project_links" links={this.state.formFields.project_links}/>
-        </div>
+        <LinkList elementid="project_links" links={this.state.formFields.project_links}/>
 
         <div className="form-group">
           <label>Files</label>
--- a/common/components/common/projects/EditProjectForm.jsx
+++ b/common/components/common/projects/EditProjectForm.jsx
@@ -155,26 +155,25 @@ class EditProjectForm extends React.PureComponent<Props,State> {
           <ImageUploadFormElement form_id="project_thumbnail_location" currentImage={this.state.formFields.project_thumbnail}/>
         </div>
         
-        <h2 className="form-group subheader">DETAILS</h2>
         <div className="form-group">
-          <label htmlFor="project_name">Project Name</label>
+          <label htmlFor="project_name"><b>Project Name</b></label>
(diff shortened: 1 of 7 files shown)


common/components/common/projects/EditProjectForm.jsx
```diff
@@ -155,7 +155,6 @@ class EditProjectForm extends React.PureComponent<Props,State> {
           <ImageUploadFormElement form_id="project_thumbnail_location" currentImage={this.state.formFields.project_thumbnail}/>
```
marlonkeating: The thumbnail upload button looks better as a standard button than the new '+' button, so we'll want to update the css here.

StanislavLev:
I updated "FileUploadButton" it seems more easy than css, but let me know if there are some problems with this way.

common/components/common/upload/FileUploadButton.jsx
```diff
@@ -29,12 +30,29 @@ class FileUploadButton extends React.PureComponent<Props, State> {
   }
   
   render(): React$Node {
-    return (
-      <div>
-        <input type="button" value={this.props.buttonText} onClick={this._handleClick.bind(this)} />
-        <input ref="fileInput" type="file" style={{display:"none"}} accept={this.props.acceptedFileTypes} onChange={this._handleFileSelection.bind(this)} />
-      </div>
-    );
+    if(this.props.thumbnail_image == "true"){
```
marlonkeating: I think a better way to structure this component would be to have a buttonText prop for where we want a regular button, and a glyphIcon prop (taking in 'fa-plus' in this case) for where we want an icon.

A couple other points:
-Be sure to add new props to the Props type up above
-It's best to pass around the boolean constant true, not the string "true".  And use the === comparison operator on.

common/components/forms/ImageUploadFormElement.jsx
```diff
@@ -48,7 +48,7 @@ class ImageUploadFormElement extends React.PureComponent<Props,State> {
           ? this._renderThumbnail()
           : this._renderThumbnailPlaceholder()
         }
-        <FileUploadButton acceptedFileTypes="image/*" buttonText="Upload Project Image" onFileUpload={this._handleFileSelection.bind(this)}/>
+        <FileUploadButton acceptedFileTypes="image/*" buttonText="Upload Project Image" thumbnail_image="true" onFileUpload={this._handleFileSelection.bind(this)}/>
```
marlonkeating: For future reference, use {true} for passing boolean values as props.

StanislavLev:
Added new prop to <FileUploadButton> and changed conditions for rendering.

common/components/common/upload/FileUploadButton.jsx
```diff
@@ -12,7 +13,8 @@ export type FileUploadData = {|
 type Props = {|
   onFileUpload: (FileUploadData) => void,
   buttonText: string,
-  acceptedFileTypes: string
+  acceptedFileTypes: string,
+  faIconClass: string
```
marlonkeating: I would name this just iconClass, since we could easily swap this out with a glyph library other than FontAwesome.

common/components/common/upload/FileUploadButton.jsx
```diff
@@ -29,12 +31,29 @@ class FileUploadButton extends React.PureComponent<Props, State> {
...
-        <input type="button" value={this.props.buttonText} onClick={this._handleClick.bind(this)} />
-        <input ref="fileInput" type="file" style={{display:"none"}} accept={this.props.acceptedFileTypes} onChange={this._handleFileSelection.bind(this)} />
-      </div>
-    );
+    if(this.props.faIconClass && this.props.buttonText){
+      return (
+        <div>
+          <input ref="fileInput" type="file" style={{display:"none"}} accept={this.props.acceptedFileTypes} onChange={this._handleFileSelection.bind(this)} />
+
+          <label>{this.props.buttonText} &nbsp;
```
marlonkeating: Make header text non-clickable for now.

common/components/forms/LinkList.jsx
```diff
@@ -98,15 +98,15 @@ class LinkList extends React.PureComponent<Props,State>  {
       <div>
         <input type="hidden" ref="hiddenFormField" id={this.props.elementid} name={this.props.elementid}/>
         
-        {this._renderLinks()}
-        
-        <Button
-            bsStyle="primary"
-            bsSize="large"
+        <label>Project Links &nbsp;
```
marlonkeating: Make header text non-clickable for now.

common/components/forms/PositionList.jsx
```diff
@@ -98,16 +98,16 @@ class PositionList extends React.PureComponent<Props,State>  {
     return (
       <div>
         <input type="hidden" ref="hiddenFormField" id={this.props.elementid} name={this.props.elementid}/>
-        
-        {this._renderPositions()}
-        
-        <Button
-            bsStyle="primary"
-            bsSize="large"
+        <label>Open Positions &nbsp;
```
marlonkeating: Make header text non-clickable for now.

PR #53, opened 2018-07-12
Project Specific Links

PeterBreen:
This PR adds four fields for specific links we want "called out" as part of a project
 - Code Repository(e.g. github)
 - Messaging (e.g. Slack)
 - Project Management (e.g. Trello)
 - File Repository (e.g. Google Drive)

The links are stored in the same backend as any other link added by the project owners, but are named specifically to be pulled out for editing/viewing separately.

Intended behavior is as follows:
when creating a project, there are four fields (for the listed four categories above) - putting a link in any of those boxes should save to the project_links array just like using the Add link button in the form and then save into the backend correctly. When viewing the "About" page for a project the name should display as human readable; e.g. "Code Repository" instead of "link_coderepo."When editing a previously created project, these links should pre-populate the correct boxes in the form, and should save correctly if modified or updated.

My tests show it all does that most times, but not all times. I'm not sure what's causing the intermittent problems. I'd like some specific review on repeatability of success/failure so I can narrow down possible faults.


PeterBreen:
Updated with requested changes, merged latest changes from master.

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -102,50 +115,87 @@ class EditProjectForm extends React.PureComponent<Props,State> {
...
-  
+
   onSubmit(): void {
     //Sanitize project url if necessary
     if(this.state.formFields.project_url) {
       this.state.formFields.project_url = url.appendHttpIfMissingProtocol(this.state.formFields.project_url);
     }
-    this.forceUpdate();
+   // create input array
+   var eLinks = ['link_coderepo','link_messaging','link_filerepo','link_projmanage'].map(name => ({linkName: name, linkUrl: this.state.formFields[name]}))
```
marlonkeating: To condense further, you can use linkNames.keys().map(...)

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -102,50 +115,87 @@ class EditProjectForm extends React.PureComponent<Props,State> {
...
+    var combinedArray = this.state.formFields.project_links.concat(eLinksArray);
+    // setState new combined array
+    this.setState({ formFields: { project_links: combinedArray }});
   }
-  
+
+  filterSpecificLinks(array) {
+    //this function updates the entire state.formFields object at once
+    var specificLinks = _.remove(array, function(n) {
+      return n.linkName.startsWith("link_")
```
marlonkeating: To make it more clear, you can replace this with
return n.linkName in linkNames;

PR #57, opened 2018-07-23
June release

marlonkeating:
Additions:
-Forgot Password link now on the bottom of Login page

Fixes/Tweaks:
-Allow selecting countries other than US on Alert signup modal
-Erase fields on Alert signup after submitting (useful when we have the site up as a signup 'kiosk' at conventions and the like)
-Change id on Alert signup email field so that it will not autopopulate with emails from other sites (mainly for the kiosk scenario)
-Single-select Tag Fields on the Project create/edit page (like Issue Area and Organization) can now be cleared

### DIFF
--- a/civictechprojects/static/css/LogInController.css
+++ b/civictechprojects/static/css/LogInController.css
@@ -33,3 +33,8 @@
   background-color: #dddddd;
   cursor: no-drop;
 }
+
+.LogInController-forgotPassword {
+  text-align: center;
+  font-size: 11px;
+}
--- a/civictechprojects/forms.py
+++ b/civictechprojects/forms.py
@@ -94,17 +94,9 @@ class ProjectCreationForm(ModelForm):
         project.project_name = form.data.get('project_name')
         project.project_url = form.data.get('project_url')
 
-        issue_areas = form.data.get('project_issue_area')
-        if issue_areas and len(issue_areas) != 0:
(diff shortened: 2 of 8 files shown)


civictechprojects/static/css/LogInController.css
```diff
@@ -33,3 +33,8 @@
   background-color: #dddddd;
   cursor: no-drop;
 }
+
+.LogInController-forgotPassword {
+  text-align: center;
+  font-size: 11px;
+}
```
PeterBreen: `text-align: center;` doesn't seem to be needed here.
 
I'd suggest raising the font size. 11px is smaller than anything else on the page, it's small enough I have readability concerns, and it's going to be a tough mobile clicktarget. 

One option would be to move it inline with but aligned right in the `Password:` div above the password box, which would let us keep it at 16px (default) size but also keep it out of the way for people who don't need it.
Something like this would work:
```
<div style="clear: both;">
  <span style="float: left;">Password:</span>
  <span style="float: right;"><a href="#">Forgot password?</a></span>
</div>
```
(But without inline styles.)

PeterBreen:
Password reset link moved works, zip codes not required for non-US countries works, signup alert when search yields no results works. Looks good.

I know we talked about the password reset token situation in standup but I'll put it here too: I'd like some way to expire tokens and communicate that to the user on the front end.

PR #58, opened 2018-07-25
Project page UI - Renaming labels and sorting positions alphabetically

andy-edvalson:


common/components/controllers/AboutProjectController.jsx
```diff
@@ -228,29 +228,30 @@ class AboutProjectController extends React.PureComponent<{||}, State> {
...
-          {
-            position.descriptionUrl
-            ? <a href={position.descriptionUrl} target="_blank" rel="noopener noreferrer">{positionDisplay}</a>
-            : <span>{positionDisplay}</span>
-          }
-          </div>
-        );
-    });
+    return project && project.project_positions && project.project_positions
+      .sort(function(a, b) {
```
marlonkeating: You should be able to do the sorting in one line using
https://lodash.com/docs/4.17.10#sortBy

Ex:
_.sortBy(project.project_positions, [function(a) { return (a.roleTag.subcategory + a.roleTag.display_name).toLowerCase(); }])

common/components/controllers/AboutProjectController.jsx
```diff
@@ -228,29 +228,21 @@ class AboutProjectController extends React.PureComponent<{||}, State> {
...
-          <div key={i}>
-          {
-            position.descriptionUrl
-            ? <a href={position.descriptionUrl} target="_blank" rel="noopener noreferrer">{positionDisplay}</a>
-            : <span>{positionDisplay}</span>
-          }
-          </div>
-        );
-    });
+    return project && project.project_positions && _.chain(project.project_positions).sortBy('roleTag.tag_name').value()
```
marlonkeating: tag_name is a unique identifier for a tag that doesn't generally factor in the subcategory.  We'll want to sort by the subcategory + display name.

PR #62, opened 2018-07-27
Reset Filters button

PeterBreen:
Adds a "Reset Filters" button to remove both saved search query and all selected filter tags in one click.

Known issues:
Clicking Reset Filters will remove the search string and tags from the tag container, but does not update the querystring / rerender the page. It requires clicking twice to actually re-render projects. I'd like some eyes on this and the querystring code to see where the problem is.

The component pulls props from ProjectTagContainer parent component which is fine _if_ it stays in this component, but that may not always be true. I'd rather it work no matter where it is based on the fluxStore rather than passing state as props, but my attempts at that weren't successful.

Style/positioning isn't quite right, I'll need to go back and do a proper pass on that when function is fixed up. On the other hand I think I can remove the ProjectTagContainer conditional which adds or removes the `.ProjectTagContainer-noTags` class, based on some refactoring I did for the TagContainer's base class. That needs further testing.


civictechprojects/static/css/ProjectSearchBar.css
```diff
@@ -32,4 +32,4 @@
 button.ProjectSearchBar-submit:disabled {
   background-color: #ededed;
   color: #7a7a7a;
-}
\ No newline at end of file
+}
```
marlonkeating: Any changes in this file?

PeterBreen: No, I had some .css there before I moved the button into the TagContainer instead of the search bar. There shouldn't be any changes to this file at all. Maybe Atom likes adding/removing newlines?

PeterBreen:
Wrote a 'CLEAR_FILTERS' dispatcher and callback function. It just resets the state.keyword and state.tags to an empty string and an empty List(), respectively. If this looks like I'm on the right path here I'll look into why the callback isn't triggering a page render when it ought to, otherwise let me know if I should change the _clearFilters() callback.

common/components/componentsBySection/FindProjects/ProjectTagContainer.jsx
```diff
@@ -38,6 +38,7 @@ class ProjectTagContainer extends React.Component<{||}, State> {
             tag => <ProjectTag key={tag.tag_name} tag={tag}/>,
           )
         }
+        <ResetSearchButton tags={this.state.tags}/>
```
marlonkeating: Is the tags prop still being used?

common/components/stores/ProjectSearchStore.js
```diff
@@ -85,6 +87,8 @@ class ProjectSearchStore extends ReduceStore<State> {
         return this._loadProjects(state);
       case 'SET_KEYWORD':
         return this._loadProjects(this._addKeywordToState(state, action.keyword));
+      case 'CLEAR_FILTERS':
+        return this._clearFilters(state);
```
marlonkeating: Wrap this line in _loadProjects to refresh the page contents

PeterBreen:
Thanks for the catch on that unneeded props and wrapping the callback in loadProjects. It seems to refresh for me correctly now, so this is function complete unless you want me to also work on making the Reset Filters button only display conditionally (i.e. only show up if a keyword or tag is active)

PR #66, opened 2018-08-06
Add Project Stage field to edit project

PeterBreen:
Adds Project Stage field to create/edit project page, consisting of a selectable dropdown allowing for project owners to select where they are in the overall process from idea to finished product.

I've got one bug and two questions: 

- editing Technologies Used, e.g. adding a programming language or database, causes an error. This is new to the branch, master doesn't have it, but I'm not sure where it comes from. I get this error:
   ```
   TagSelector.jsx:75 Uncaught TypeError: selectedValues.map is not a function
       at TagSelector.handleSelection (TagSelector.jsx:75)
   ```
     I didn't change anything related to the Technology selector so ??? on this one.

And two questions I'd like resolved before I finalize the PR for review:

- In the dropdown, Project Stage items sort alphabetically. Do we want them to sort in terms of completion, i.e. Ideation at the top, Finished at the bottom of the list? Or is this ok as is?

- On the About Project page, where should the project stage be displayed? At the top in the same box as Title, Website, Issue Area, etc? Or elsewhere?

### DIFF
--- a/common/components/controllers/AboutProjectController.jsx
+++ b/common/components/controllers/AboutProjectController.jsx
@@ -75,12 +75,12 @@ class AboutProjectController extends React.PureComponent<{||}, State> {
                     </div>
                     <div className="row">
                       <div className="col">
-                        {project && !_.isEmpty(project.project_issue_area) && project.project_issue_area[0].display_name}
+                        {project && !_.isEmpty(project.project_issue_area) ? "Issue Area: " + project.project_issue_area[0].display_name : null}
                       </div>
                     </div>
                     <div className="row">
                       <div className="col">
-                        {project && !_.isEmpty(project.project_stage) && project.project_stage[0].display_name}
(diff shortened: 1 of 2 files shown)


common/components/utils/ProjectAPIUtils.js
```diff
@@ -29,6 +29,7 @@ export type ProjectAPIData = {|
   +project_id: number,
   +project_description: string,
   +project_issue_area: $ReadOnlyArray<TagDefinition>,
+  +project_state: $ReadOnlyArray<TagDefinition>,
```
marlonkeating: Should be called 'project_stage' instead of 'project_state'.

PeterBreen:
I've got the project stage sorting correctly in the dropdown and displaying in the AboutProjectController, but still getting the TypeError error in the Technology selector. I've been looking through the call stack trying to debug it and I'm not sure what's going on with that component but it seems to be a problem bringing it out of the backend.

PR #67, opened 2018-08-07
Add floating footer to all pages, with configurable header links

marlonkeating:
Adapted the floating footer on the (currently unused) Landing page and applied it to the rest of the site's pages, with links that can be set using an environment variable.

### DIFF
--- a/civictechprojects/static/css/MainFooter.css
+++ b/civictechprojects/static/css/MainFooter.css
@@ -19,7 +19,7 @@
 }
 
 .MainFooter-footer-link::after {
-  content: " . ";
+  padding-right: 30px;
 }
 .MainFooter-footer-link:last-of-type::after {
   content: "";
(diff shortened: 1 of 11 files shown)


civictechprojects/static/css/MainFooter.css
```diff
@@ -19,7 +19,7 @@
 }
 
 .MainFooter-footer-link::after {
```
PeterBreen: If we're not using dots to separate items, this could be refactored to remove the ::after psuedoselector.

Probably something like 
```css
.MainFoooter-footer-link {
  padding: 0px 15px;
}
```
and then delete the entry for `.MainFooter-footer-link:last-of-type::after entirely. Also I'd suggest 15/15 left and right instead of 0/30 so the clicktargets are evenly balanced.

PR #69, opened 2018-08-09
npm update/install latest dependencies

PeterBreen:
Related to solving an issue I ran into building #66 this should update project dependencies to the latest versions from an `npm install` `npm update` and `npm audit` series of commands.

It should be noted `npm audit` did turn up some items which require manual intervention to fix, a potential longer term issue.

Anyway, to test: 
 - make sure no functionality of the site is altered or broken
 - specifically make sure the bundle.js built with this branch does not exhibit the errors I was getting in #66 previously


PeterBreen:
Obviously since this branch is resolving problems I'm having locally I don't think I'm the best source to test, but I've done a few things.

On prod-mirror, I've tested these and they work:
 - keyword search, tag filter, keyword/tag combined. Reset filters button.
 - View individual project pages
 
Locally on the branch, I've tested these and they work:
 - Login with existing account
 - Create, edit project (including adding and removing information)

Things that don't work for me:
 - Delete a project. Tested locally, returns 500 error.

Haven't tested the create a user or forgotten password actions.

PR #63, opened 2018-07-27
Change link behavior in 'MainHeader', remove repetition in browser history

StanislavLev:
Please check if this is a right direction :)

PR #70, opened 2018-08-10
Pre-release Fixes 8/9

marlonkeating:
This contains a number of miscellaneous fixes and additions. 

- Project links open in new tab
- Added favicon
- Added footer links to the header as well
- Migrated imgur links to AWS cloudfront
- Create Project button now always shows and just redirects to login page when user isn't logged in
- Project Description field has new placeholder text, and a configurable example link
- Added environment variable for auto-redirecting http to https
- Added Seattle Tech For Good and Data for Good organizations


common/components/componentsBySection/FindProjects/ProjectCard.jsx
```diff
@@ -28,7 +28,9 @@ class ProjectCard extends React.PureComponent<Props> {
     return (
       <a style={styles}
         className="ProjectCard-root"
-        href={url.section(Section.AboutProject, {id: this.props.project.id})}>
+        href={url.section(Section.AboutProject, {id: this.props.project.id})}
```
PeterBreen: I'd like this to be revisted by design after the double history entry bug currently open in #63 is solved. Unless we have specific testing otherwise, [this sums up my position](https://medium.com/the-metric/links-should-open-in-the-same-window-447da3ae59ba)

common/components/chrome/MainHeader.jsx
```diff
@@ -14,7 +16,7 @@ class MainHeader extends React.PureComponent<{||}> {
         <span onClick={this._onHomeButtonClick}>
           <img
             className="MainHeader-logo"
-            src="https://i.imgur.com/NhoAjjN.png"
+            src={cdn.image("dl_logo.png")}
```
PeterBreen: Just an aside here but this a 110kb 2040x612 image being displayed at 200x54. Would be worth some filesize and loadtime saving to upload a native 200x54 copy.

common/components/componentsBySection/FindProjects/ProjectCard.jsx
marlonkeating: It was a team decision, so I'd bring this up with design.

common/components/chrome/MainHeader.jsx
marlonkeating: Thanks for pointing this out.  I've updated the cdn image.

PR #76, opened 2018-08-20
display message when user attempts to sign in with wrong password

pconte:


democracylab/views.py
```diff
@@ -18,7 +18,7 @@ def login_view(request):
             login(request, user)
             return redirect('/')
         else:
-            # TODO: Show error message if the email/password combination are invalid
+            messages.error(request, 'username or password not correct')
```
marlonkeating: To keep consistent with the wordage elsewhere on the site, maybe we can start with something like "Incorrect Email or Password"

pconte:
Message text has been updated.

PR #77, opened 2018-08-22
Sort projects by last updated date

pconte:
add date_modified field to project model; create migration to add new field to db; display date_modified field on AboutProject view


civictechprojects/models.py
```diff
@@ -48,6 +48,7 @@ class Project(models.Model):
     project_name = models.CharField(max_length=200)
     project_url = models.CharField(max_length=2083, blank=True)
     project_links = models.CharField(max_length=5000, blank=True)
+    project_date_modified = models.DateTimeField(auto_now=True)
```
marlonkeating: What value will project_date_modified be in existing projects that haven't been updated?

pconte: I believe that when a model is first created, it will be treated as having been "updated", since the "auto_now" attribute specifies that the date field will be set when the model is saved.  Is that correct?

package-lock.json
```diff
@@ -10,7 +10,7 @@
       "integrity": "sha1-zgBCgEX7t9XrwOp7+DV4nxU2arI=",
       "requires": {
         "@babel/types": "7.0.0-beta.51",
-        "lodash": "^4.17.5"
+        "lodash": "4.17.10"
```
marlonkeating: Did we need to update the packages here?

pconte: I ran npm run install a while back, and I was keeping the package-lock.json out of the commits manually, but not reverting it locally.  But I got lazy today and just did "add .", and figured it was okay to to push it up.  I will go research the changes in those lodash versions, though, to be sure there aren't any potential breaking changes.

pconte:
I brought in the moment.js library and a React component that wraps moment.js.  The last updated date is now rendered inside a Moment tag.  I also moved the last updated date down to the subtext areas along with Issue Area and Location.

pconte:
Sorry for taking so long to get to the last issue, re: date changed being the next day.  When hydrating the project model to json, the date_modified datetime object was being converted to a string using a formatter, so the front-end was receiving a formatted datetime string which corrupted the timezone information.  I modified the hydrate-to-json method to use the __str__() method on the date_modified field, so the raw string is returned unformatted with all information intact.  That includes the time in UTC format.  The front-end is smart enough to take the UTC time and convert it to the local timezone of the browser.

PR #78, opened 2018-08-29
Remove empty filters and filter categories from DOM

PeterBreen:
This PR is composed of two major parts. The first, in views.py and helpers/projects.py, uses the tag model and the projects model to create a new `num_times` value passed to the front end with every API request for a category of tags. This num_times value represents the number of projects which have that tag, i.e. how many projects will display to the user if a given filter is checked.

On the front end, num_times is used in two ways. Simplest is that this value is shown to the user: If Python is tagged in 5 projects, the Python filter will read `Python (5)`. 

It is also used to remove empty filters from the filter component entirely. Some previous commits in this branch removed empty filters fairly late in the process of generating the filters so empty categories would show up where I didn't want them - they were initialized before I removed empty filters. By changing where I remove empty filters to the very start of SelectorCollapsible's `initializeOptions` function this problem is avoided.

To test: 
load Find Projects page and expand all filter categories. All categories shown should have at least one entry in them. All entries shown should have a parenthetical number of >=1. If an entry reads 5, selecting that filter should return 5 projects.

Adding or removing a project's tags should result in numbers updating and filters appearing/disappearing as appropriate.

*Known Issue:*
Filtering via a saved querystring, i.e. pasting localhost:8000/index/?section=FindProjects&issues=public-safety into your browser and submitting, does not work. I get an exception error in console from the flux project store. Any ideas on that?

### DIFF
--- a/civictechprojects/views.py
+++ b/civictechprojects/views.py
@@ -9,7 +9,7 @@ from time import time
 from urllib import parse as urlparse
 import simplejson as json
 from django.views.decorators.csrf import csrf_exempt
-from django.db.models import Q
+from django.db.models import Q, Count
 from .models import Project, ProjectFile, FileCategory, ProjectLink, ProjectPosition, UserAlert
 from .helpers.projects import projects_tag_counts
 from common.helpers.s3 import presign_s3_upload, user_has_permission_for_s3_file, delete_s3_file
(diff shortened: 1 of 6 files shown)


civictechprojects/views.py
```diff
@@ -18,20 +18,27 @@
 from democracylab.models import Contributor, get_request_contributor
 from common.models.tags import Tag
 
-from pprint import pprint
-
 def tags(request):
     url_parts = request.GET.urlencode()
     query_terms = urlparse.parse_qs(
         url_parts, keep_blank_values=0, strict_parsing=0)
     if 'category' in query_terms:
```
marlonkeating: Add a TODO for adding a flag to add the tag counts, as there are some instances (like populating the project tag dropdowns) where we don't need the tag counts.

PR #79, opened 2018-08-30
fix create/edit project errors

PeterBreen:
This PR fixes some issues that stemmed from #78.  

- Change views.py to output in a consistent format (list). I accidentally made it put out a dictionary in one case and a list in the other in #78. I also refactored the front end filter conditional to continue to work properly after this change.
 - views.py now has a `getCounts` boolean conditional to allow for asking for tag counts or not, so we can choose only to take the perf hit when needed instead of always. This also means the ProjectAPIUtils request function has been changed to accept true/false as an argument for getCounts and each instance of an API request in the project now has a true or false argument as appropriate. **this could use some review** I think.
 - Modifies edit project initialize function to explicitly wait for setState to complete - this seems to be fixes towards, but not completely fixing `this.state.displayList is undefined` errors. All I can say is the number of times I've seen the error has gone down. See [this commit message](https://github.com/DemocracyLab/CivicTechExchange/commit/3926b7dca2e705f03049d6d6de9bcadad75389d5#r30390112) for specifics. 

Basically this branch fixes some mistakes I made in #78 and hopefully the getCounts true/false filter will help clear up some performance problems by only taking that perf hit when needed.

The reason I ask for review in that second bullet point: I'm not confident that I made the changes in ProjectAPIUtils and views.py the best way. I feel like it should have better default handling both in ProjectAPIUtils (if the argument isn't passed into the function, default to false) and on the Python backend (if the argument isn't in the querystring, default to false again) but I'm not sure how best to go about that. 

For testing:

- The filters on the find project page will pass `getCounts=true` and get the filtered results, every other instance of an API request by category will pass false and not get those. Everything related to the user flow for tag categories should work correctly: View projects, edit project, create project.
- Editing projects should work consistently (more consistently) and not return undefined errors in the console.

### Known Issues
 - Does not fix the querystring bug referenced earlier in #78 because I've been told another PR has a fix for it.


PeterBreen:
Updated comment with latest info on bugs. Edit workflow still not working 100% - will keep working on it.

PeterBreen:
After our discussion on this I implemented database caching on this. Right now it's a basic 5 minute timeout which is meant to be tested on prod-mirror or staging to see if it has any useful performance benefit; if that works I'll go back and make it cache intelligently.

democracylab/settings.py
```diff
@@ -99,6 +99,14 @@
 
 LOGIN_REDIRECT_URL = '/'
 
+#Caching number of tag counts only for now - change this if other things are db-cached
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.db.DatabaseCache',
+        'LOCATION': 'active_tag_counts',
```
marlonkeating: We may want to use this cache table for other things, so let's name it something more generic.

democracylab/settings.py
```diff
@@ -99,6 +99,14 @@
 
 LOGIN_REDIRECT_URL = '/'
 
+#Caching number of tag counts only for now - change this if other things are db-cached
```
PeterBreen: Copy/pasted the wrong thing into my commit message. It should read this: May need to run `python manage.py createcachetable` per the [documentation](https://docs.djangoproject.com/en/2.1/topics/cache/#database-caching).

civictechprojects/views.py
```diff
@@ -22,7 +22,7 @@
 
 #TODO: Set getCounts to default to false if it's not passed? Or some hardening against malformed API requests
 
-@cache_page(300) #cache duration in seconds, cache_page docs: https://docs.djangoproject.com/en/2.1/topics/cache/#the-per-view-cache
+@cache_page(1200) #cache duration in seconds, cache_page docs: https://docs.djangoproject.com/en/2.1/topics/cache/#the-per-view-cache
```
PeterBreen: When we get a task scheduler of some type this should be set to the scheduler interval, I just wanted a longer value for more testing.

PeterBreen:
Okay, I _believe_ I've closed the loop here. 

The latest commits add a command script to refresh the API endpoint cache when run manually or with a scheduler. This branch requires the requests library to be installed: `pip install requests` 
Once installed, run the script like so: `python manage.py updatecache`

 I'm not sure of a few things, however:

1) Is it clearing the cache successfully before it starts? If not, it's not going to successfully regenerate new cache values. I'm not getting errors, but I'm also not sure how to test this.
2) requests.get() seems to require a hardcoded domain or else it gives a schema error; I couldn't give it a relative URL of `/api/tags/etc` for example. Therefore, to **test locally** be sure to change the domain variable in updatecache.py to `http://127.0.0.1:8000` (or whatever your dev environment is) -- but ideally there's some way to determine domain name dynamically?

common/components/common/selection/SelectorCollapsible.jsx
```diff
@@ -65,13 +65,9 @@ class SelectorCollapsible<T> extends React.PureComponent<Props<T>, State> {
...
     if(!_.isEmpty(props.options)) {
-
-      let filteredOptions = Object.keys(props.options).filter(function(key) {
-        return props.options[key]['num_times'] > 0;
-      }).map(function(key) {
-        return props.options[key]
-      });
-
+      let filteredOptions = props.options.filter(function(key) {
+          return key.num_times > 0;
```
pconte: Should we filter out tags with a count of zero?  I guess it reduces clutter/noise in the filter list.  But unless/until more tags are connected to active projects, the smaller subset of initially selected tags could associate the DemocracyLab brand with those arbitrarily selected tags (e.g. could unintentionally pigeonhole D.L. is being only about PHP and MySQL if that is all that the user sees in the available filter list).

PeterBreen: Yes, that's intentional/intended on my part. This is solving a UX problem with the live site right now where empty filters are causing bad user experience, and based on some conversations I had with some people on the team, this is the best "get it fixed now" solution - and I'd argue it's a solid contender for "get it fixed period" but if it needs to be extended to meet that second goal more thoroughly, that can happen in the future.

PeterBreen:
Cache update script has a domain using settings value instead of a hardcoded one. I've done some testing locally and the cached counts of tags updates as expected (when script is run they update, otherwise they stay cached until timeout) for both adding and removing tags.

civictechprojects/management/commands/updatecache.py
```diff
@@ -0,0 +1,24 @@
+# Scheduled tasks for Heroku Scheduler (or similar)
+# currently, regenerates the views.py cache_page for tags(request)
+from django.core.cache import cache
+from django.core.management.base import BaseCommand, CommandError
+from django.conf import settings
+import requests #requires pip install requests
```
marlonkeating: Can omit this comment, as users can see it in requirements.txt

civictechprojects/management/commands/updatecache.py
```diff
@@ -0,0 +1,24 @@
...
+class Command(BaseCommand):
+    help = 'Refreshes API endpoint cache'
+
+    def handle(self, *args, **options):
+        #clear old cache values first - note that this will wipe the ENTIRE cache, if we start caching multiple views or other files this will need to be specifically wiping just the API endpoint cache
+        cache.clear()
+        #define endpoints
+        endpoints = ['Issue(s) Addressed', 'Technologies Used', 'Role', 'Organization', 'Project Stage']
+        #define URL parts for loop.
+        #TODO: Get domain dynamically, not hardcoded.
```
marlonkeating: Can remove this comment now

PR #82, opened 2018-09-05
Filter by location

pconte:
TODO:
1) Would like to set the location and sort dropdown values if the URL initially has query param values (e.g. if page is refreshed, or if user clicks on a link containing query params)

QUESTION:
2) Should the user be able to clear a location or sort filter by choosing the "-----" option?  Right now that option is disabled, so the user cannot clear either filter.  If we make that option enabled, should we change the copy?

### DIFF
--- a/common/components/stores/ProjectSearchStore.js
+++ b/common/components/stores/ProjectSearchStore.js
@@ -190,6 +190,7 @@ class ProjectSearchStore extends ReduceStore<State> {
   _clearFilters(state: State): State {
     state = state.set('keyword', '');
     state = state.set('sortField', '');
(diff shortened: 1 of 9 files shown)


common/components/stores/ProjectSearchStore.js
```diff
@@ -172,6 +182,11 @@ class ProjectSearchStore extends ReduceStore<State> {
     return state;
   }
 
+  _addLocationToState(state: State, location: string): State {
+    state = state.set('location', location);
+    return state;
+  }
+
   _clearFilters(state: State): State {
```
marlonkeating: Let's also clear the location here.

PR #83, opened 2018-09-12
Expanded User Profile

marlonkeating:
This adds a number of new entries to the User model to support Volunteer use cases.  To test, click on the new 'My Profile' link, and fill out your volunteer information.


common/components/controllers/AboutUserController.jsx
```diff
@@ -0,0 +1,156 @@
...
+    return this.state.user ? this._renderDetails() : <div>Loading...</div>
+  }
+
+  _renderDetails(): React$Node {
+    const user: UserAPIData = this.state.user;
+    return (
+      <div className="AboutProjectController-root">
+        <div className="container-fluid">
+          <div style={{backgroundColor: "white"}}>
+            <div className="row" style={{margin: "30px 0 0 0", padding: "10px 0"}}>
```
pconte: Do you have an opinion on using inline styles like this?  In this instance, I actually like it, since this kind of box styling is really local to the component, and creating a class seems like an unnecessary abstraction.

marlonkeating: We do it both ways in the code, and generally leave it up to the authors to decide which way makes sense.

PR #85, opened 2018-09-14
Bug fixes for 9/13 release

marlonkeating:
Fixing a couple bugs that were found during testing:

- Navigating to url with sort/location arguments weren't setting the dropdowns correctly
- Saved resumes weren't being loaded into the correct file selector widget on the Edit Profile page


pconte:
React-select looks good. And thanks for putting that decode on the location param.

PR #87, opened 2018-09-22
Cypress alpha

pconte:
I'm learning cypress for the first time, and I have to say, I am impressed so far.  In this PR, I have written a couple simple tests around logging in to the DemocracyLab site.  The test uses a fixture (a collection of users, each with username and password attributes).

I used the following pages as guides:
1. To help me get the code up and running: https://blog.bitsrc.io/testing-react-apps-with-cypress-658bc482678
2. Tutorials I watched the past few days: https://docs.cypress.io/examples/examples/tutorials.html#4-Loading-data-with-fixtures
3. A recipe for a login test that I used as a template: https://github.com/cypress-io/cypress-example-recipes/tree/master/examples/logging-in__html-web-forms
4. More recipes here: https://docs.cypress.io/examples/examples/recipes.html

I installed a new package via yarn, called cypress.  I also added a couple commands to package.json.

You will first need to run "yarn install" so that the cypress package installs.  Then run the python server, so that the site is running locally.  Then type "yarn cypress" which will open the Cypress UI.  Click on the "Run tests" button, and a browser window should open.  Within that browser window, there is a left pane that shows the status of each test.  In the middle pane is the live site.

If you run the command "yarn cypress:all", the tests will run headlessly without opening the UI, and screenshots and video will be saved.

Right now, my user in the fixture is based on a user that actually exists in my local database.  So these tests should not pass for you.  We'll have to think of a way to seed the database with users, by somehow hooking into the Django ORM.  I'm also not sure if seeding users like that will work if the authentication-over-email step us skipped.  Can we passed in an auth code directly during the seeding process?


pconte:
Here's a possible solution for seeding the database prior to running E2E tests.  This article explains how to use Knex.js to seed a postgresql database with JSON data: https://medium.com/@jaeger.rob/seed-knex-postgresql-database-with-json-data-3677c6e7c9bc ...   The only real work will be writing the Knex.js seed code, which will specify any normalized relationships.  And then maintaining that code, if needed, each time a new feature is added to the codebase.  The JSON can be pulled from Cypress fixture files, so there will be one source of truth for the data in both the database and in the test assertions.

pconte:
Instead of using Knex.js (and likely duplicating all that normalized relationship code that already exists in the Django ORM code), we could use manage.py loaddata.  From this article https://coderwall.com/p/mvsoyg/django-dumpdata-and-loaddata it looks like loaddata wants each JSON object's attributes to be wrapped inside of a "fields" collection (based on what the results of running the reciprocal dumpdata command look like), inline with additional metadata attributes.  I'm imagining we could dump data from a current DL database, curate it into proper test data, and then use it with a manage.py loaddata command that can be accessed within cypress via a cy.exec('manage.py loaddata <file>') call.  This might be a better approach, since it works within the Django ecosystem.  Any potential pitfalls I am missing?

pconte:
This most recent commit is an attempt to seed the local database using the manage.py loaddata command.  I managed to get the command to execute via cypress, but the command fails.

PeterBreen:
Closing due to age; if this issue is still relevant please feel free to reopen.

PR #89, opened 2018-09-23
install lodash-webpack-plugin and babel-plugin-lodash; modify webpack config

pconte:
I followed the simple instructions here: https://github.com/lodash/lodash-webpack-plugin.  I skipped modifying the rules, but I did modify the plugins object.  Superficially, what these plugins do is transform the lodash JS code so that certain parts of the lodash library are not needed, and then excludes those parts of the lodash library from the build.  The one caveat is that the entire app will need regression testing to ensure this didn't break anything.

For the curious, I did some analysis, and lodash is used in 23 JS and JSX files, and 22 different lodash methods are referenced (mostly isEmpty and assign).  After installing the plugins and modifying webpack.config, "npm run build" produced a bundle.js of 1.1MB (down from 3.8MB, a significant reduction).


marlonkeating:
There have been a number of changes to the webpack build setup since this work was done.  It's not clear these changes are needed now, and in any case would need to be adjusted to fit our current setup.  Feel free to reopen if you would like to revisit this issue.

PR #90, opened 2018-09-25
SASS Implementation

PeterBreen:
This branch implements SASS for CSS precompiling for performance, consistency, and future growth.

SASS was chosen over LESS based on a comparative test of both; node-sass (which binds to lib-sass) is commonly used, well supported, and in my opinion, integrates with the existing DemocracyLab toolchain better than LESS. Thanks to @pconte for help on researching this.

### Why use a preprocessor?
 - Performance: Instead of loading 25 distinct CSS files with 25 individual HTTP GET requests, this will allow us to load one.  Based on testing of democracylab page load, CSS is second only to the bundle.js in terms of slowing down initial page load, and #88 and #89 address bundle related performance.
 - Consistency: Variables can help keep common style declarations exactly with a lower chance for error, and make future redesigns easier and faster. 
 - Future growth: Cleaning up and using partials makes any future changes (adding/removing Bootstrap components, using per-component CSS partials) easier to manage. 

### Using SASS
First, `npm i node-sass` or an `npm install` with the updated package.json should get you ready with the pre-requisites.
The pre-existing `npm run build` has been modified and will compile a .css file and move it to the appropriate location. So just run that like usual.

The CSS has been rearranged like so: All existing .css files, which were effectively partials, have been moved into a /partials/ subdirectory and renamed _*filename*.scss. All third-party CSS files, for example react-select, were moved into a /vendor/ subdirectory and similarly renamed. _styles.scss_ is the main SASS file, which is used to define which partials get put into the build in which order. __vars.scss_ is a special partial for all variable assignments. When compiled, the output file is placed in /compiled/styles.css.

If any new partials are needed, create them in the appropriate subdirectory (partials for us, vendor for third-party) and name them `_filename.scss`, then add an `import` statement to styles.scss to include them in the build, for example`@import 'partials/mynewpartial';` (the underscore in the filename tells SASS it's a partial but it's not required in the import - `@import partials/name` and `@import partials/_name` both work)

### PR To do:
Before this PR is ready to be merged, issues to be addressed:
 - [x] review all existing CSS files, evaluate as partials to make sure they're in use and that the cascade is correct
 - [ ] build out `_vars.scss` with common values found in the above review as a starter use of variables, expand in the future
 - [x] investigate vendor files, such as react-select and bootstrap css; they're a big part of the overall load and stripping them down (if possible) would be good
    - [x] related to the above, integrate whatever part or the whole of Bootstrap into /vendor/ instead of going to MaxCDN.
- [x] integrate npm run sass into the npm run build process, so it doesn't require two steps (probably Webpack)
- [ ] document SASS in the DL Getting Started guide, for current and future developers
- [ ] On the server side, reconcile Chrome Audit showing cache TTL as 1 minute against real CSS cache time (and if it is 1 minute, make it longer!)

Optionally,
-  rename the output to styles.min.css or something, to make it clearer which file is which. styles.css and styles.scss could be confusing.
- update the section of the build script which copies files into the correct location for the live site to not copy over the new .scss partials, they don't hurt but just take up space.
- ensure the old .css files are removed from the server, as with the partials they just take up space

This is my idea of an MVP; I just want to make a platform for that growth while addressing currently present issues. Any ideas on that are great.

also I really like sass, it's good. :)

https://sass-lang.com/
https://www.npmjs.com/package/node-sass


pconte:
Re: your first TODO in your list, I really liked component-scoped CSS.  Vue.js makes this a snap, but React has a few ways to achieve it, as well.  One possibility is for us to use CSS Modules.  There are things that we can do with webpack to transform the CSS to component-scoped CSS.  I also like the idea of importing the stylesheets into the component, and accessing the styles that way.  At that point, the styles can be transformed to JS within the component, and some nice things can be done with styling logic.

PeterBreen:
> Re: your first TODO in your list, ... 

I think that's a good suggestion but I'd consider it one for a future PR. Right now, first and foremost, I'm trying to solve the specific loadtime performance issues we're having in a quick, minimally-invasive way; there's no major architecture changes to how the CSS is applied, only in how it's compiled. I consider this an iterative improvement -- if we choose another approach in the future it won't be any more difficult than now, but we get performance benefits now. And if we don't choose another approach, there's a lot of potential for future growth.

PeterBreen:
I'm reconsidering my current approach on trying to write a custom "these are all the bootstrap style declarations we use" css file. The puppeteer script I'm using has some limitations: Any conditionally applied style (e.g. using psuedoselectors) isn't picked up so it's either fix the script or reconsider the approach. 

I should start how I want people to go on: In the future if someone uses Bootstrap I proposed that the best way to do it would be to import just the function they need and just the styles they need rather than just importing the whole thing - so why not do the same here?

https://getbootstrap.com/docs/4.0/getting-started/theming/ suggests this should be straightforward. Rather than getting Bootstrap.min.css from their CDN as a precompiled file and trying to strip out pieces, I can download their source files and import the partials we actually use. This should integrate seamlessly with the sass implemention in this branch and strike a good balance between reducing our CSS payload, actually following the practice I want to follow, and not driving any other developer up a wall when they try to reuse any existing Bootstrap classes or components.

pconte:
Module- and component-based bootstrap is the way to go.

civictechprojects/static/css/styles.scss
```diff
@@ -22,6 +23,7 @@
 @import 'partials/LandingController';
 @import 'partials/MainHeader';
 @import 'partials/MainFooter';
+@import 'partials/project_signup'; // has a modal opacity rule required for bootstrap-modals?
```
PeterBreen: I see that we use the react-bootstrap fade class to make this fade in but in my experience it doesn't actually fade in, it just appears. May be worth revisiting this if we extend our use of modals.

package.json
```diff
@@ -7,8 +7,8 @@
   "author": "Marlon Keating <marlonakeating@gmail.com>",
   "license": "MIT",
   "scripts": {
-    "build": "webpack --config webpack.config.js && python manage.py collectstatic --noinput",
-    "sass": "node-sass civictechprojects/static/css/styles.scss -o civictechprojects/static/css/compiled --output-style compressed",
+    "build": "webpack --config webpack.config.js && npm run build:sass && python manage.py collectstatic --noinput",
```
PeterBreen: I read a lot of conflicting positions on whether to use webpack or npm scripts to build the .css from sass partials and then if webpack, _how_ to configure it. While I investigate that situation further I modified the npm run build script to compile the css for now so it's one step instead of two and requires no changes to the developer workflow in terms of commands run.

civictechprojects/static/css/_vars.scss
```diff
@@ -0,0 +1,17 @@
+//pattern for variable names: generic-to-specific.
+//when possible name variables after function
+
+//colors
+$color-orange: #f79e02; // brand
+$color-green: #86c232;
+$color-green-dark: #61892f;
```
pconte: At some point, we will want to move away from variable names the imply certain values (like "green") and use names that specify the styling role (like "primary-button-color", or like "color-text-light").  That's not for right now, though, since we do not have those styling roles specified.  Ideally, a designer would determine what those styling roles are based on patterns in the visual components.

PR #95, opened 2018-10-10
display error message when existing user tries to sign up

pconte:


democracylab/views.py
```diff
@@ -48,6 +48,7 @@ def signup(request):
         else:
             # TODO inform client of form invalidity
             print('Invalid form', form.errors.as_json())
+            messages.error(request, form.errors.as_json(escape_html=True))
```
pconte: The form (or combination of the form and Django ORM) already appears to be catching the error when an existing user tries to sign up.  As an interim solution, I just sent the form errors to the flash box at the top of the form.  I noticed that the form errors in an object, and it looks like the intention is that it is used to decorate the field where the error occurred.  This is really how we want to code up our forms.  However, that is not how they work right now.  For the time being, we can just send this block of errors to the flash box.  In order for this PR to be ready-to-merge, I still have to format the form errors output, as right now it is just raw JSON.

PR #98, opened 2018-10-16
Bootstrap fix

PeterBreen:
My mistake here. In one of my sass branches I accidentally left the MaxCDN Bootstrap CSS include intact and that made it into the PR - so we've been loading Bootstrap twice (once from MaxCDN, and once from the compiled styles.css locally) - not helpful, to say the least.

Aside from performance, it also means there may be issues with the site style that haven't been uncovered yet because they were temporarily hidden by the maxCDN bootstrap 'filling in.' My check of code coverage on load suggests there won't be much, so the small details are likely to be where it happens.

I've removed the MaxCDN stylesheet from the new_index.html template and done a smoke test of editing, creating, and deleting a project as well as editing a user profile, so the core loop works. I ran into some layout problems, solved by adding the bootstrap forms and grid partials.  I'll keep testing more thoroughly before I flag this as ready for review, but I wanted it up asap.


PeterBreen:
I've done some testing on prod mirror and I think this is ready for review and merge. If there are any further missing Bootstrap issues they should be fast to fix if identified.

PR #102, opened 2018-11-06
Card Update

PeterBreen:
This PR consists of two halves, updating the _card UI_ and updating the _card data._

## Card UI
The card UI should match the UI given by design and function at all screen breakpoints. The project page is now controlled by Bootstrap's container class and uses Bootstrap's flexbox-based row/column classes to facilitate this.

## Card Data
The project cards should return all project data correctly including previously-used fields such as project title but also the newly included fields of project website, skills needed, and description. All three of these fields  currently use truncation to compromise between fitting as much information on the card as possible with keeping cards reasonably sized; this truncation should work as follows:

  - description: [X] character limit until truncated with "..."
  - skills needed: [Y] skill limit before the list is truncated with Y-1 entries and a "Z more" entry, where Z is the number of skills a project needs not shown here (Should be "total - 3" at this point)
  - project website: Should not display http:// or https:// prefix.

If any field does not reach the truncation limits it should display unmodified. 

#### General
- [x] look into some small card border spacing to keep the logos slightly out of the corners and fix the description hitting the right edge of the card
- [x] if there's no website, don't display the globe in subinfo (actually for all 3 items do this)
- [x] some skills like "Design: All" has a display name of "All" - planned to be fixed in another branch for this release
- [x]  truncation limits are hardcoded instead of being in state 
- [x] subinfo icons don't line up cleanly when stacked vertically (slightly different widths)

#### 2 wide cards
- [x] uneven card heights when cards are displayed 2-wide. `height: calc(100% - 16px)` on the card container works

#### 1 wide cards
- [x] Grid layout on short cards (i.e. a not-very-tall logo) has some overlapping issues with skills needed and description. Setting title to grid-row 1, description to grid-row 2, skills to grid-row 3 and leaving logo and subinfo alone seemed to fix this in my browser devtools
- [x] the globe/pin/clock can, if the data is very long, sit on a line by itself with a full linebreak before the actual info happens. This might happen in 2 wide too?

#### related page content
- [x] remove the `0.375rem 0.75rem` padding values from the date modified and location sorts in the filters; makes them less constrained at tablet/narrow desktop resolutions
- [x]  The nested Project card container-fluid class should be `.container-fluid pl-0 pr-0` to correctly use 100% of its available width, padding/spacing is provided by the parent container
- [x]  Bootstrap row/col-12 classes for intro text, search and filter bar need to be checked, so page content linebreaks properly.

#### Feedback from team meeting: 
- [x] add XL breakpoint for maximized desktop view (2 'wide' cards, matches original Figma spec here)
- [x] change Skills Needed "None" to "Contact Project for Details"
- [x] (stretch goal) look at skills needed truncation, maybe return "...X more" instead of "..."
- [x] add vertical grey line (on right edge of logo extending full height) at tablet/XL widths

#### XL breakpoint
- [x] give subinfo some top spacing so logos don't press so closely, check subinfo ul margin-bottom

#### optional stretch goals:
- navigation and header/footer content should be constrained in width with page content, but keep the background color of the bars 100% browser width like they are now. (Generally "should be fixed for mobile")


PeterBreen:
Fixed card height at 2 wide, truncation is passed as a prop, made changes to text grey and truncation limits as proposals for design review.

PeterBreen:
Fixes issues and requests brought up by review this morning, including an XXL breakpoint, resolving the Skills Needed: None question, and I made a change to the skills needed truncation to try to ease a UI issue which showed up some times where "..." was the only 'skill' on a line, which looked odd. Also resolved an outstanding issue I noticed with card padding.

PeterBreen:
Added vertical grey line between logo/subinfo and the rest of the card at appropriate breakpoints (sm to lg, xl+)

PeterBreen:
There's a small visual error with the grey line _only_ at the largest viewport due to how the 2-wide card heights are equalized. Every other breakpoint either doesn't have the problem or doesn't render the line due to the card layout being different. Otherwise all the issues I know of are solved, ready for review.

common/components/componentsBySection/FindProjects/ProjectCard.jsx
```diff
@@ -28,88 +25,85 @@ class ProjectCard extends React.PureComponent<Props> {
...
+    <div className="ProjectCard-skills">
+    <h3>Skills Needed</h3>
+      {this._generateSkillList(this.props.skillslen)}
+    </div>
+    )
+  }
+  _generateSkillList(numskills): React$Node {
+    //take array of skills needed from props, truncate if required, and map to list items
+    const skills = Truncate.arrayT(this.props.project.positions, numskills)
+    let index = 0; //super hacky unique key indexes goooo
```
marlonkeating: You can also get the index from the second parameter in the map function.
`skills.map((skills,i) => <li key={i}>{skills}</li>)`

common/components/componentsBySection/FindProjects/ProjectCard.jsx
```diff
@@ -28,88 +25,85 @@ class ProjectCard extends React.PureComponent<Props> {
...
+    );
+  }
+  _renderSubInfo(): React$Node {
+    //only renders a list item for ones where we have data, otherwise skip
+    return (
+      <div className="ProjectCard-subinfo">
+        <ul>
+        {this.props.project.location &&
+          <li>
+          <i className="fas fa-map-marker-alt fa-fw"></i>{this.props.project.location}
```
marlonkeating: Maybe indent and put icon and text on separate lines for readability.

common/components/utils/ProjectAPIUtils.js
```diff
@@ -33,7 +33,9 @@ export type ProjectAPIData = {|
   +project_location: string,
   +project_name: string,
   +project_thumbnail: FileInfo,
-  +project_claimed: boolean
+  +project_claimed: boolean,
```
marlonkeating: Didn't we get rid of project_claimed?

common/components/utils/ProjectAPIUtils.js
```diff
@@ -72,10 +74,22 @@ class ProjectAPIUtils {
       name: apiData.project_name,
       thumbnail: apiData.project_thumbnail,
       claimed: apiData.project_claimed,
-      date_modified: apiData.project_date_modified
+      date_modified: apiData.project_date_modified,
+      url: apiData.project_url,
+      positions: apiData.project_positions && apiData.project_positions.length !=0
```
marlonkeating: The null/empty check can be covered with _.isEmpty()

common/components/utils/ProjectAPIUtils.js
```diff
@@ -72,10 +74,22 @@ class ProjectAPIUtils {
...
+          ? ProjectAPIUtils.getSkillNames(apiData.project_positions)
+          : ['Contact Project for Details'],
     };
   }
 
+  static getSkillNames(positions: array) {
+    let output = [];
+    positions.forEach(function(data) {
```
marlonkeating: Could be done in a single line with .map

common/components/utils/truncate.js
```diff
@@ -0,0 +1,24 @@
+// @flow
+
+class Truncate {
+
+  static stringT(str: string, length: number) {
+    var dots = str.length > length ? "..." : "";
+    return str.substring(0, length) + dots;
```
marlonkeating: To fulfill the function's implied contract we should include the length of the ellipsis.
`return (str + dots).substring(0, length);`

common/components/utils/truncate.js
```diff
@@ -0,0 +1,24 @@
...
+  static arrayT(arr: array, length: number) {
+    if (arr.length > length) {
+     let result = arr.slice(0, length-1)
+     result.push(arr.length - 3 + " more")
+     return result;
+    } else {
+      return arr;
+    }
+  }
+  static urlPrefixT(url: string) {
```
marlonkeating: nit: I wouldn't call this a truncation so much as a beautification.  url.js might be a more appropriate place to put it.

common/components/utils/ProjectAPIUtils.js
PeterBreen: Do you want me to apply this to the other checks in the same function, issue area and project stage?

marlonkeating: If it's code you wrote as part of this change, sure.  Otherwise, I'd say let's not expand the scope of this large pr.

common/components/utils/truncate.js
PeterBreen: In my testing this truncates off the ellipsis, so I'll need to rethink it. I agree that we should include the length of the ellipsis in the returned string, however.

common/components/utils/ProjectAPIUtils.js
PeterBreen: It's not, so I won't expand scope.

common/components/utils/truncate.js
marlonkeating: Yeah, now that I think about it, maybe something like this:
`return str.substring(0, length - dots.length) + dots;`

PR #104, opened 2018-11-14
Implement Volunteer management workflow

marlonkeating:
Volunteers can now apply to join projects, and project owners have the ability to manage their volunteers.

- User can apply to join by clicking the 'Volunteer with Project' button on the About Project page
- Project owners will receive email when users apply to join, and they have the option to approve or reject the volunteer on the About Project Page
- Approved volunteers are shown on a project's About page
- My Projects page shows projects user is volunteering with


civictechprojects/models.py
```diff
@@ -374,3 +370,45 @@ def create_or_update(email, filters, country, postal_code):
...
+    is_approved = models.BooleanField(default=False)
+    projected_end_date = models.DateTimeField(auto_now=False, null=True)
+
+    def __str__(self):
+        return 'Project: ' + str(self.project.project_name) + ', User: ' + str(self.volunteer.email)
+
+    def to_json(self):
+        volunteer = self.volunteer
+
+        # TODO: Get user
```
PeterBreen: Does this volunteer_json fulfill the TODO?

common/components/common/FeedbackModal.jsx
```diff
@@ -0,0 +1,72 @@
...
+                <div className="character-count">
+                  { (this.state.feedbackText || "").length} / {this.props.maxCharacterCount}
+                </div>
+                <textarea className="form-control" rows="4" maxLength={this.props.maxCharacterCount} value={this.state.feedbackText}
+                          onChange={this.onTextChange.bind(this)}>
+                </textarea>
+              </Modal.Body>
+              <Modal.Footer>
+                  <Button onClick={this.confirm.bind(this, false)}>Cancel</Button>
+                  <Button onClick={this.confirm.bind(this, true)}>{this.props.confirmButtonText}</Button>
```
PeterBreen: Might be worth using one of the [button classes](https://getbootstrap.com/docs/4.0/components/buttons/) on the confirm. `className="btn btn-success"` possibly. Same for other confirm action modals in the PR (e.g. Contact Project, Volunteer with Project) if they don't inherit from this. We should be consistent about it in modals sitewide and we're not so it may merit its own issue instead of fixing on this branch - up to you.

common/components/common/projects/ProjectVolunteerModal.jsx
```diff
@@ -0,0 +1,165 @@
...
+  roleTag: TagDefinition,
+  showConfirmationModal: boolean
+|};
+
+const volunteerPeriodsInDays: $ReadOnlyArray<SelectOption> = [
+  ["Less than 1 week",7],
+  ["1 week - 1 month",30],
+  ["1 month - 3 months",90],
+  ["3 months - 6 months",180],
+  ["6 months - 1 year",365]
```
PeterBreen: Maybe account for all possible durations by adding an open-ended "greater than a year" option?

common/components/common/projects/ProjectVolunteerModal.jsx
```diff
@@ -0,0 +1,165 @@
...
+    return (
+      <div>
+        <ConfirmationModal 
+          showModal={this.state.showConfirmationModal}
+          message="Do you want to apply to this project?"
+          onSelection={this.receiveSendConfirmation}
+        />
+          <Modal show={this.state.showModal}
+                 onHide={this.closeModal}
+                 style={{paddingTop:'20%'}}
```
PeterBreen: I think we don't need this inline padding after #99 was merged in.

common/components/common/projects/ProjectVolunteerModal.jsx
marlonkeating: On the database side we're storing the end date, so the question in that case would be what to make the end date.  I think we can revisit this if we find out people want to commit to more than a year.

common/components/common/FeedbackModal.jsx
marlonkeating: Yeah, that should probably be a wider fix across the site.

PR #105, opened 2018-11-15
Project short description field

marlonkeating:
Added a 140 character short description field for projects, which will now be displayed on the project card if it exists.  If the short description field isn't present, we will fall back to a truncated version of the long description.


common/components/controllers/AboutProjectController.jsx
```diff
@@ -129,6 +129,7 @@ class AboutProjectController extends React.PureComponent<{||}, State> {
             <div className="col">
               <h2 className="form-group subheader">PROJECT DETAILS</h2>
               <div className="Text-section" style={{whiteSpace: "pre-wrap"}}>
+                {project && project.project_short_description ? project.project_short_description + "\n\n" : null}
```
PeterBreen: I think we could use a separate heading here, matching with the form field labels. So a "Short Description" and a "Project Description", while keeping them both under the Project Details H2 like they are now. Right now with a live project where you have both it's not clear which is which.

civictechprojects/models.py
```diff
@@ -94,7 +96,7 @@ def hydrate_to_tile_json(self):
             'project_id': self.id,
             'project_name': self.project_name,
             'project_creator': self.project_creator.id,
-            'project_description': self.project_description,
+            'project_description': self.project_short_description if self.project_short_description else self.project_description,
```
PeterBreen: I like this. I was envisioning bringing them both out to the React frontend and picking there but this is nicer. 👍

common/components/controllers/AboutProjectController.jsx
marlonkeating: I put them under separate 'PROJECT SUMMARY' and 'PROJECT DETAILS' for now.

PR #106, opened 2018-11-15
Filter category "Select All" options

PeterBreen:
Create "Select All" checkboxes for each category of filters. If there are subcategories (such as Roles Needed > Business") only display a checkbox in the subcategories.

Further details: https://trello.com/c/1dG5uXbJ/112-add-filtering-by-tag-category

Currently multi-tag adding works in a sense ... they're correctly collected and sent to the proper component but we need to refactor the store's ADD_TAG to accept multiple tags at once, as sending 8 or so ADD_TAGs at once overwhelms it in a bad way.

- [x] place checkboxes in appropriate location
- [x] collect correct info for add and remove operations
- [x] send to/remove from flux store in one operation
- [ ] select all checkboxes interact properly with their category/subcategory


common/components/common/selection/SelectorCollapsible.jsx
```diff
@@ -160,20 +162,24 @@ class SelectorCollapsible<T> extends React.PureComponent<Props<T>, State> {
...
-  _selectAll(category): React$Node {
-    //get list of tags
-      category.map(function(tag) {
-       return tag.tag_name
-    })
-    //find a way to access list of selected tags in TagSelectorCollapsible
-    // if tag in the above map isn't selected, select it, and then check the Select All box
-    // otherwise remove any tags from the above map from list of selected tags and uncheck the box
+
+  _handleSelectAll(category): React$Node {
```
PeterBreen: For background, the parent component is `TagSelectorCollapsible` which passes a lot of props to `SelectorCollapsible`  -- I'm trying to work sort of in parallel to how the tag selection/unselection works as is.

This code, which doesn't work, is supposed to respond to someone clicking the select all box by first checking whether or not it's checked (looking at this.state.isAllChecked) and if the box wasn't checked, run the code to check all the boxes in a given category. If it was checked (and the user is unchecking it) run the code to uncheck all the boxes in a given category.

This is a little tricky because `this.selectOption` (which goes up to `SelectOption()` in TagSelectorCollapsible) will automatically check an unchecked box or uncheck a checked one - but in this case since I want to check everything or uncheck everything I have to filter what I provide to that function, otherwise someone with half the boxes checked in a category would just flip which half was selected/unselected.

Well, that's the theory anyway.

PeterBreen:
Just pushing up a progress commit, still have some issues to resolve.
- [ ] adding all with select all works (in Issue Areas, the only place I'm testing right now) and so does remove all ... once. After that I get `state.tags is undefined` errors. Need to figure out what's happening in ProjectSearchStore.
- [ ] in SelectorCollapsible, updateAllState needs to actually tie checkbox state to the, well, state of the checkboxes in a given category as opposed to a disconnected true/false flip. Probably need to solve the first one.
- [ ] make sure it works in subcategories (like Roles Needed -> Software Development) because I had some questions there

common/components/common/tags/TagSelectorCollapsible.jsx
```diff
@@ -57,21 +58,35 @@ class TagSelectorCollapsible extends React.Component<Props, State> {
...
-      metrics.logSearchFilterByTagEvent(tag);
-    } else {
       ProjectSearchDispatcher.dispatch({
-        type: 'REMOVE_TAG',
-        tag: tag,
+        type: opts.type,
+        tag: tagnames,
       });
+    } else {
+      metrics.logSearchFilterByTagEvent(tag);
```
PeterBreen: I think I've got this in the right place but we'll need to review metrics for the add-many-tags event in any case

common/components/common/tags/TagSelectorCollapsible.jsx
```diff
@@ -94,6 +111,24 @@ class TagSelectorCollapsible extends React.Component<Props, State> {
     );
   }
 
+  _AllChecked(): void {
```
PeterBreen: Here's the big problem function right now. Aside from the name, which I should probably make better, what I want is something that will check or uncheck the Select All checkbox for a given category of filters based on whether or not all the filters are themselves checked.

so two main issues I've got right now - it's not updating correctly in the DOM so I need to fix that, and I also think it's going to run into problems with subcategories due to how those components are structured.

common/components/stores/ProjectSearchStore.js
```diff
@@ -171,10 +179,19 @@ class ProjectSearchStore extends ReduceStore<State> {
   }
 
   _addTagToState(state: State, tag: string): State {
+    //tag does not have to be a string, this works for an array of strings as in ADD_MANY_TAGS
     const newTags:$ReadOnlyArray<string> = state.tags.concat(tag);
     return state.set('tags', newTags);
   }
 
+  _removeTagsFromState(state: State, tags: array): State {
```
PeterBreen: I was happy to find the existing ADD_TAG function worked just fine to add multiple tags at once instead of just one, but I haven't tested the REMOVE_TAG for the same. Worth a look just so we can keep our code clean.

common/components/common/selection/SelectorCollapsible.jsx
```diff
@@ -142,14 +145,44 @@ class SelectorCollapsible<T> extends React.PureComponent<Props<T>, State> {
...
+    //this feels like a mess, but find what tags are selected or not and then either select or unselect the remainder
+    if(this.props.isAllChecked) {
+      let toUpdate = category.filter(tag => this.props.optionEnabled(tag) === true)
+      this.selectOption(toUpdate, {multiple: true, type: "REMOVE_MANY_TAGS"} );
+    } else {
+      let toUpdate = category.filter(tag => this.props.optionEnabled(tag) === false)
+      this.selectOption(toUpdate, {multiple: true, type: "ADD_TAG"} );
+    }
+    this.props.AllChecked();
```
PeterBreen: instead of calling this manually I think I need it to be part of an onchange event, right now it's always one interaction behind. if you manually start checking items one at a time and then look at the console you'll see for example it will say items checked 1 when you've just checked the 2nd item in a category

common/components/common/tags/TagSelectorCollapsible.jsx
```diff
@@ -100,13 +99,23 @@ class TagSelectorCollapsible extends React.Component<Props, State> {
...
               onOptionSelect={this.selectTag.bind(this)}
+              selectedByCategory={this._selectedByCategory()}
             />
             )
           : null
         }
       </div>
     );
   }
+  _selectedByCategory(): void {
```
PeterBreen: This generates selected item counts, which seems to work okay, and is what I planned to use to figure out if any given Select All checkbox should be checked, along the lines of "I'm the select all checkbox for the Data subcategory, Data has a length of 2 items, and this.props.selectedByCategory has data = 2, check the box"

common/components/common/selection/SelectorCollapsible.jsx
```diff
@@ -225,7 +227,7 @@ class SelectorCollapsible<T> extends React.PureComponent<Props<T>, State> {
       : 0;
     this.setState({chevronX});
   }
-  //
+  // this is from when I had the state controlling checkboxes in TagSelectorCollapsible, for ref
```
PeterBreen: reference code that should be deleted, oops

common/components/common/selection/SelectorCollapsible.jsx
```diff
@@ -157,17 +157,17 @@ class SelectorCollapsible<T> extends React.PureComponent<Props<T>, State> {
...
+    //it works because we can rely on the _.groupBy() and filteredOptions in this.initializeOptions, idx 0 of any given cat/subcat is guaranteed to exist (no empty categories allowed) and return the right value due to grouping.
     let selectAllId = category[0].subcategory || category[0].category
     return (
       <span>
         <label className="CollapsibleMenuItem">
           <
             input type="checkbox"
             className="ContextualCollapsible-selectAll"
-            checked={this.state[selectAllId]}
+            checked={this._selectAllController.bind(this, category)}
```
PeterBreen: checked={} and onChange={} are where this all goes poorly. _handleSelectAll works more or less assuming I feed it the right conditional value to check (it's presently wrong) and the _selectAllController is uhhhh well I'd just delete it and start over

PeterBreen:
Closing this PR for now since it's not likely to get finished before we redevelop the find projects page anyway which is going to remove the need for it.

PR #107, opened 2018-11-16
Update glyph definitions

marlonkeating:
We recently switched to FontAwesome version 5.5, but didn't update certain font definitions that had been changed/deprecated.  This change updates those definitions, centralizes them in a single index, and also cleans up some old django artifacts and UI issues.


common/components/controllers/AboutProjectController.jsx
```diff
@@ -212,7 +213,7 @@ class AboutProjectController extends React.PureComponent<{||}, State> {
   _renderProjectLocation(): React$Node {
     if(this.state.project && this.state.project.project_location && (this.state.project.project_location !== Locations.OTHER)) {
       return <div className="col">
-        <i className="fa fa-map-marker fa-1" aria-hidden="true"></i>
+        <i className={GlyphStyles.MapMarker} aria-hidden="true"/>
```
PeterBreen: `<i>` elements aren't self-closing, for spec compliance they need the `</i>` closing tag.

PR #111, opened 2018-11-20
Khoa/hack to give thanks

khuynh92:
- [x] Make the Login/Signup page responsive
- [x] Add a left-side drawer and hamburger in mobile view
- [x] Make the main header and sub header responsive
- [x] Make the 'my projects' page responsive


common/components/controllers/AboutProjectController.jsx
```diff
@@ -117,6 +119,7 @@ class AboutProjectController extends React.PureComponent<{||}, State> {
               ? <div className="row" style={{margin: "30px 40px 0 40px"}}>
                   <div className='col'>
                     <h2 className="form-group subheader">TECHNOLOGIES USED</h2>
+                    <Button>hi</Button>
```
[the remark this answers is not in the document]
khuynh92: This was a test, I will remove this on my next push to this branch

civictechprojects/static/css/partials/_MainHeader.scss
```diff
@@ -21,11 +25,43 @@
...
+}
+
+.MainHeader-navDrawer {
+  margin-top: 25px;
+  width: 43.72vw;
+  min-width: 200px;
+}
+
+.MainHeader-drawerSpan {
+  margin-top: 10px;
```
PeterBreen: inline elements like span don't take top/bottom margins, only left/right, and it does feel like the hamburger menu items are pretty tightly grouped on a phone as a result. I'd like to see the clicktargets be larger and the items be spaced a little further apart vertically to make tapping with a finger easier. 

Changing these to block elements (a div, most likely) and using padding instead of margin should resolve the spacing and the clicktarget size.

khuynh92: Thanks for reviewing my commits. I'll make sure to make these changes within the next few days

PeterBreen:
Tested both locally and on prod-mirror; looks good and works well. Approved from my end.

civictechprojects/static/css/partials/_SubHeader.scss
```diff
@@ -11,22 +11,84 @@
...
+  margin: auto;
   padding-right: 40px;
 }
 
 .SubHeader-rightContent {
-  float: right;
+  display: flex; 
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
```
PeterBreen: One quick thing I missed in my review, sorry! Could you set the justify-content so it only applies `center` on mobile? On desktop resolutions it should be the default `flex-start` - the nav menu looks odd for a logged out user otherwise.

civictechprojects/static/css/partials/_SubHeader.scss
```diff
@@ -11,22 +11,84 @@
...
+  height:100%;
+    float: right;
+}
+
+.SubHeader-leftContent {
+    float: left;
+    display: flex; 
+    flex-direction: column;
+    align-items: start;
+    justify-content: center;
```
PeterBreen: One quick thing I missed in my review, sorry! Could you set the justify-content on LeftContent so it only applies `center` on mobile? On desktop resolutions it should be the default `flex-start` - the nav menu looks odd for a logged out user otherwise.

khuynh92: Alright, I fixed it!

PR #117, opened 2018-12-05
Add project owner(s) to project profile page

TCW417:
Updated model to include more owner info (based on VolunteerUserData type).  Updated front end to show (for now) single project owner just above volunteer section.  Owner passed in an array, allowing for future handling more than one owner.


civictechprojects/models.py
```diff
@@ -402,6 +403,37 @@ def to_json(self):
...
+        relation = VolunteerRelation()
+        relation.project = project
+        relation.volunteer = volunteer
+        relation.projected_end_date = projected_end_date
+        relation.application_text = application_text
+        relation.save()
+
+        relation.role.add(role)
+
+class OwnerRelation(models.Model):
```
marlonkeating: I don't see this code being used in the views yet, nor a django migration file to add the required tables to the database.  If it's not going to be used for this iteration, we should at least comment it out for now.

Though even then, I wonder if it might make more sense to just add an 'owner' flag to the existing VolunteerRelation, since we will likely be allowing volunteers to be 'promoted' to owners.

TCW417: Agreed. Using the existing VolunteerRelation class and just adding a property would be easier than creating an entirely new infrastructure.  This particular python, I think, I started on then change my plans and just didn't catch needing to remove it.

PR #124, opened 2019-01-03
Update on volunteer actions

TCW417:
Added calls to save project, forcing update to last updated date, whenever volunteers are approved, rejected, promoted or demoted.  Also added a metrics.logVolunteerPromoted function and called it from the appropriate function so as to be consistent with the other volunteer API functions.


common/components/utils/metrics.js
```diff
@@ -137,6 +137,10 @@ class metrics {
     _logEvent("ProjectDismissVolunteer", {userId: userId, projectId: projectId});
   }
   
+  static logProjectPromoteVolunteer(userId: number, projectId, number): void {
```
marlonkeating:
```suggestion
  static logProjectPromoteVolunteer(userId: number, projectId: number): void {
```

civictechprojects/views.py
```diff
@@ -338,10 +338,12 @@ def volunteer_with_project(request, project_id):
 @csrf_exempt
 def accept_project_volunteer(request, application_id):
     volunteer_relation = VolunteerRelation.objects.get(id=application_id)
+    project = volunteer_relation.project
```
marlonkeating: I'd move this code to a helper method to specifically update the project timestamp.  It's not clear from the context why exactly we're saving the project all over the place.

TCW417: Moved to volunteer_relations.update_project_timestamp() helper method.

PR #136, opened 2019-01-22
Khoa/navbar update

khuynh92:
- [x] Followed design requirements according to the design team
- [x] Created donate button
- [x] Link Donate button to correct donation site
- [x] Dynamically show Icon if no user photo exists
- [x] on Icon click, show profile links
- [x] added drawer/hamburger to SubHeader
- [x] refactor and reorganize code

**notes**
- I began merging all the components from MainHeader to SubHeader, since SubHeader had the basic layout already. Should we rename all classes in SubHeader to MainHeader?
- Currently the header works, but could be refactored to be cleaner. Next , I will continue to clean up the code, and possibly create new component for the Drawer/Slider


common/components/chrome/SubHeader.jsx
```diff
@@ -43,26 +61,228 @@ class SubHeader extends React.Component<{||}, State> {
...
+              <a href="mailto:hello@democracylab.org" target="_blank" rel="noopener noreferrer">
+                <div className={'SubHeader-drawerDiv'} onClick={() => this._navigateToSection('FindProjects')}>
+                  Contact Us
+                </div>
+              </a>
+              <Divider />
+              {
+                !CurrentUser.isLoggedIn() &&
+                <React.Fragment>
+                  <div className={'SubHeader-drawerDiv'} onClick={() => this._navigateToSection('LogIn')}>
```
PeterBreen: There's a `this._onLogInClick` function used for the not-mobile login button, is there a reason not to reuse it here?

common/components/configs/SectionLinkConfigs.js
```diff
@@ -3,17 +3,17 @@ import Section from '../enums/Section.js'
 const SectionLinkConfigs = [
   {
     section: Section.FindProjects,
-    title: 'FIND PROJECTS',
+    title: 'Find Projects',
```
PeterBreen: Thank you for doing this bit of cleanup. Good to do, if we want all caps in the DOM we can use text-transform CSS rules instead. More readable code = better. :)

common/components/controllers/MainController.jsx
```diff
@@ -34,7 +34,7 @@ class MainController extends React.Component<{||}> {
 
   render(): React$Node {
     return [
-      <MainHeader key='main_header'/>,
+      // <MainHeader key='main_header'/>,
```
PeterBreen: Can remove this to clean up the render function if we don't need it - I know you mentioned merging MainHeader and SubHeader in the PR. I'll check with Marlon but I don't see a problem having a single header component considering how simple our header is.

common/components/chrome/SubHeader.jsx
```diff
@@ -170,7 +170,7 @@ class SubHeader extends React.Component<{||}, State > {
   };
 
   _closeDropdown(event): void {
-    if (this.anchorEl.contains(event.target)) {
+    if (this.anchorEl && this.anchorEl.contains(event.target)) {
```
PeterBreen: I added this to the conditional because I was getting an undefined error in the following situation: On mobile, log out and go to the Find Projects page. Open the hamburger menu and click Find Projects (aka the page you're on) - I don't think this is the best way to go about it since I think it triggers a page reload (something that checks if page requested = current page and just closes the slider if true would be ideal?) but it was quick.

common/components/chrome/SubHeader.jsx
```diff
@@ -262,23 +262,31 @@ class SubHeader extends React.Component<{||}, State > {
   }
 
   _renderHeaderLinks(): React$Node {
-    return FooterLinks.list().map((link, i) =>
-      <div className="SectionLink-root" key={i}>
-        <a className="SubHeader-anchor" href={link.url} target="_blank" rel="noopener noreferrer">
-          <h3>{link.name}</h3>
-        </a>
-      </div>
+    //TODO: Extract the links into a component much like FooterLinks
```
PeterBreen: The FooterLinks component had a Donate button added to it, so we were getting two donate buttons in the header. In the future I'd like to see this refactored to be not hardcoded links but since we only have two, I did it for the sake of releasing this week. Your code was perfectly solid, but the data it drew from changed in another PR.

common/components/chrome/SubHeader.jsx
```diff
@@ -262,23 +262,31 @@ class SubHeader extends React.Component<{||}, State > {
...
+        </div>
+      </React.Fragment>
     );
   }
 
   _renderIcon(): React$Node {
     return (
-      this.state.user && this.state.user.user_thumbnail ?
-        <img className="SubHeader-userImg" src={this.state.user.user_thumbnail} /> :
+      this.state.user && !_.isEmpty(this.state.user.user_thumbnail) ?
```
PeterBreen: I ran into an issue with a logged-in user with no profile photo; just checking on `this.state.user.user_thumbnail` wasn't robust enough since an empty object returned true and then it expected an img src that didn't exist. So I used lodash's isEmpty for a slightly more robust conditional.

PeterBreen:
Sorry about pushing up to your branch without giving you a chance to fix this stuff; but in order to code freeze today for a release this week I was asked to just push up the fixes to the issues I'd identified. Please do feel free to continue working on the code refactor/cleanup you mentioned. Thanks for all the work you did here!

common/components/chrome/MainHeader.jsx
```diff
@@ -1,226 +1,336 @@
...
       return (
-        <span key={i}>
-          <a
-            href={link.url}
-            target="_blank" rel="noopener noreferrer"
-            onClick={FooterLinks.logClick.bind(this, link)}
-          >
-            {link.name}
+        <React.Fragment>
+          <a key={link.url} href={link.url} target="_blank" rel="noopener noreferrer">
```
PeterBreen: Since we've fixed the double-entry in the history bug, I think we can stop using target="_blank" here, the back button works.

PR #139, opened 2019-01-30
Michael branch2

Lennerblom:
hid skills needed and technologies used when there are none to display


civictechprojects/static/css/partials/_AboutProject.scss
```diff
@@ -177,6 +177,9 @@
     }
 
   }
+  .AboutProjects-hide {
+    display: none;
```
marlonkeating: Using CSS to hide elements works, but the convention we've adopted for the project is to just not render hidden elements in the first place.

marlonkeating:
These changes were made in https://github.com/DemocracyLab/CivicTechExchange/pull/147

PR #142, opened 2019-02-07
Add title and description for find projects page and project profiles

marlonkeating:
Currently search engine results don't show good titles or descriptions, and this change aims to fix that.  

Also included are some open graph tags to (theoretically) allow thumbnail previews in Facebook and similar platforms.  However, this may not work as Facebook only seems to read the tags when rendered from the server.


common/components/controllers/AboutProjectController.jsx
```diff
@@ -231,6 +234,19 @@ class AboutProjectController extends React.PureComponent<{||}, State> {
...
+  
+  _renderHeader(project: ProjectDetailsAPIData): React$Node {
+    const title: string = project.project_name + " | DemocracyLab";
+    const description: string = project.project_short_description || Truncate.stringT(project.project_description, 300);
+    
+    return (
+      <Headers
+        title={title}
+        description={description}
+        thumbnailUrl={project.project_thumbnail.publicUrl}
```
PeterBreen: If no project logo is defined, I get a `Uncaught TypeError: Cannot read property 'publicUrl' of undefined` error and the page fails to load.

To duplicate: Visit any project detail page where no logo is defined, such as https://democracy-lab-prod-mirror.herokuapp.com/index/?section=AboutProject&id=114

To resolve: Do we have a default fallback image we can use in those cases - DL logo maybe? If this is for the FB or similar sharing that sounds okay to me. I ran into a similar issue with the user thumbnail conditional check on another PR and lodash isEmpty solved it for me -- see line 308 in /common/components/chrome/MainHeader.jsx.

PR #149, opened 2019-02-27
Peter/filter update

PeterBreen:
MVP of filter redesign. See [card on Trello](https://trello.com/c/xl6uNxPJ/161-story-redesign-filter-experience) for details, including tasks still to do before it's "finished," but that's probably for another sprint. 

Specifically, this is MVP function for the filter component _only_ and not the associated project page redesign to, e.g., tag box, sort and search component location and style, etc.


PeterBreen:
MVP now including the current style as per March 5 (see Figma) but no further completed to-do items. Even so, these filters should be faster to open/close, provide more information to our users, and be easier to use on both desktop and mobile.

common/components/componentsBySection/FindProjects/Filters/ProjectFilterDataContainer.jsx
```diff
@@ -0,0 +1,150 @@
...
+    );
+  }
+
+  _renderFilterCategories(): void {
+    //iterate through this.state.sortedTags into key/value pairs, one per component
+    //first get category names in an array to iterate over
+
+
+    //define the order we want our filters to be in (these must match database "category" values to be valid)
+    const fixedOrderCategories = ["Issue(s) Addressed", "Role", "Technologies Used", "Project Stage", "Organization"]
```
marlonkeating: I think we might be able to simplify the sorting logic if we use a category configuration object.  For example:
const categoryConfig = {"Issue(s) Addressed":{orderNumber: 1},"Role": {orderNumber: 2} ...}

And then we can do the sorting by doing something like
const categoryGroups = _.toPairs(this.state.sortedTags);
const orderedFilters = _.sortBy(categoryGroups , categoryGroup => tag.category in this.state.sortedTags? categoryConfig[categoryGroup[0]] : 99);

common/components/componentsBySection/FindProjects/Filters/RenderFilterCategory.jsx
```diff
@@ -0,0 +1,127 @@
...
+}
+
+//define CSS classes, keep it readable
+const classCategoryExpanded = 'ProjectFilterContainer-category ProjectFilterContainer-expanded';
+const classCategoryCollapsed = 'ProjectFilterContainer-category ProjectFilterContainer-collapsed';
+const classSubcategoryExpanded = 'ProjectFilterContainer-subcategory ProjectFilterContainer-expanded';
+const classSubcategoryCollapsed = 'ProjectFilterContainer-subcategory ProjectFilterContainer-collapsed';
+
+
+class RenderFilterCategory<T> extends React.PureComponent<Props<T>, State> {
```
marlonkeating: Should at least define what the state contains for this component.  Personally, I would put the expanded/unexpanded states into their own object within state in the event I later needed to put other values in the state.

common/components/componentsBySection/FindProjects/Filters/RenderFilterCategory.jsx
```diff
@@ -0,0 +1,127 @@
...
+            {this._renderFilterList(this.props.data)}
+          </div>
+        </div>
+    );
+  }
+
+  _renderFilterList(data) {
+    //this function renders individual clickable filter items regardless of category or subcategory status
+    let sortedTags = Object.values(data).map((tag) =>
+      <li key={tag.category + '-' + tag.display_name} className="ProjectFilterContainer-list-item">
```
marlonkeating: The expression
tag.category + '-' + tag.display_name
is repeated multiple times.  It should be a const.

PR #150, opened 2019-03-02
Refactor My Projects api and flux structure

marlonkeating:
This is preliminary work towards implementing the Re-Engage Volunteer workflow, but the refactoring is extensive enough I decided it would be good to review it on its own before proceeding.  Changes include:
- Prune fields returned by My Projects api in order to save bandwidth and database operations
- Hide My Projects button when a user owns no projects and is not volunteering with any projects
- Create a Universal Flux dispatcher to handle all dispatching, in order to support components subscribing to more than one flux store (which wasn't possible where each store uses their own dedicated dispatcher)


PeterBreen:
Prod-mirror testing doesn't have the first error with logging but does have the error of going to My Projects resulting in every edit link showing as id=undefined

I think this is it:
https://github.com/DemocracyLab/CivicTechExchange/blob/re_engage_volunteer/common/components/componentsBySection/MyProjects/MyProjectCard.jsx#L65

It's looking for `props.project.id` and I think it should look for `props.project.project_id`

PR #151, opened 2019-03-03
Standardize width for fontawesome glyphs on project card

PeterBreen:
In glyphs.js we used the `fa-fw` (fixed width) class on two of the three icons that could show on a project card on the Find Projects page, but not all three.

After reviewing the rest of the glyphs I removed the `fa-fw` class from all the glyphs and then manually added it to the project card component, letting us choose when to use fixed-width and when not to use fixed-width iconography.

See https://fontawesome.com/v4.7.0/examples/#fixed-width for documentation on fixed width icons.


common/components/componentsBySection/FindProjects/ProjectCard.jsx
```diff
@@ -87,19 +87,19 @@ class ProjectCard extends React.PureComponent<Props> {
         <ul>
         {this.props.project.location &&
           <li>
-            <i className={GlyphStyles.MapMarker}></i>
+            <i className={GlyphStyles.MapMarker + ' fa-fw'}></i>
```
marlonkeating: "fa-fw" should live in a constant, so folks know what it means.

PeterBreen: Changed to a const and named it so it's meaningful. I agree that longer term we should change the glyphs constructor to accept an options object or something like you describe.

PR #152, opened 2019-03-05
Theme button consistency fix

PeterBreen:
On certain browser/device combinations (chrome/MacOS but not chrome/Win or chrome/Linux) some of our button elements do not render the correct, themed button, instead rendering what seems like a default 'grey' button which looks disabled and we've had user issues with it.

This fix may require a few deployments for testing, but the root cause seems to be an interaction between react-bootstrap <Button>, the Bootstrap v4 styles, and our theme button override not actually overriding properly. This first step is making sure our theme button styles override correctly when building the site stylesheet, as they were not even showing up as overridden styles when generated by sass previously, suggesting the rules were being overwritten completely.

### To test:

go to: https://democracy-lab-dev.herokuapp.com

test cases:
1) go to any project page while signed _out_ and see if the “Sign in to Contact Project” and “Sign in to Volunteer” buttons are correctly themed (orange with white text)
2) go to any project page while signed _in_ and make sure the Contact Project and Volunteer with Project buttons are still correctly themed
3) Go to any project you own and make sure the “Edit Project” button is correctly themed.

Please make sure to report OS and browser, as we want to make sure it works across all of them and doesn't break where it used to work. Also note that democracy-lab-dev may not have your saved login info; if you can only verify case 1 and not 2-3 that's fine.


PeterBreen:
I'm not a big fan of !important declarations so this is mostly a fast way to see if the appearance rule is at fault. If so, and time permitting, we can refactor the declaration in Global.scss to be the proper level of specificity.

PeterBreen:
I figured out how to address the issue more cleanly, this is ready for a test deployment so I can check more browser/OS combinations than I have available in the dev environment.

Specifically, we now have a new `clear-button-appearance` helper class which can be added to any button or a[type="button"] component where the appearance declaration causes issues. This is (best I can tell) only an issue with react-bootstrap <Button> components, as regular buttons such as the navigation do not exhibit the issue, so it should be fairly uncommon.

I've also modified the theme button classes to be consistent with how we want our buttons to work on hover.

PeterBreen:
> I figured out how to address the issue more cleanly, this is ready for a test deployment so I can check more browser/OS combinations than I have available in the dev environment.
> 
> Specifically, we now have a new `clear-button-appearance` helper class which can be added to any button or a[type="button"] component where the appearance declaration causes issues. This is (best I can tell) only an issue with react-bootstrap components, as regular buttons such as the navigation do not exhibit the issue, so it should be fairly uncommon.
> 
> I've also modified the theme button classes to be consistent with how we want our buttons to work on hover.



> I figured out how to address the issue more cleanly, this is ready for a test deployment so I can check more browser/OS combinations than I have available in the dev environment.

Seems to be working as intended across browsers and OSes now, windows, mac, and linux.

civictechprojects/static/css/partials/_Global.scss
```diff
@@ -6,22 +6,26 @@ body {
...
+a.btn-theme:hover, a[type="button"].btn-theme:hover {
+  text-decoration: none;
+}
 
-  a[type="button"] {
-    color: $color-text-light;
-    text-decoration: underline;
-  }
+//use to clear specific button appearance settings, e.g. fix webkit-appearance problem on react-bootstrap <Button href=""> components
+button.clear-button-appearance, a.clear-button-appearance, a[type="button"].clear-button-apperance {
```
marlonkeating:
```suggestion
button.clear-button-appearance, a.clear-button-appearance, a[type="button"].clear-button-appearance {
```

PR #153, opened 2019-03-06
skills needed on same line as details

ashabrai:


marlonkeating:
The page design has changed considerably since this change, and this is no longer needed.

PR #154, opened 2019-03-06
Nav bar style

joyliao07:
- [x] Change top navbar background color to #FFFFFF
- [x] Top Navigation bar should always stay visible on the screen
- [x] Remove bottom navigation bar
- [x] Top right Profile image should dim, not brighten on hover-over
- [x] Mobile menu should slide in from the right, not left


civictechprojects/static/css/partials/_MainFooter.scss
```diff
@@ -1,4 +1,5 @@
 .MainFooter-footer {
+  display: none;
```
marlonkeating: Instead of hiding via CSS, we should comment out the area where we render this element in the first place.  It's the same end result, but faster since the browser can skip reading the html.

marlonkeating:
Width is looking good now, but we also need to adjust the upper padding on the rest of the site content so the floating header doesn't hide anything.

image

common/components/controllers/SectionController.jsx
```diff
@@ -22,22 +23,25 @@ import AboutUserController from "./AboutUserController.jsx";
 
 type State = {|
   section: SectionType,
+  padding: int
```
marlonkeating:
```suggestion
  padding: number
```

PR #155, opened 2019-03-06
exclude unsearchable projects from tag counters

justBanks:


civictechprojects/helpers/projects.py
```diff
@@ -3,7 +3,7 @@
 from collections import Counter
 
 def projects_tag_counts():
-    projects = Project.objects.all()
+    projects = Project.objects.filter(is_searchable='True')
```
marlonkeating: Omit the quotes, since we're testing for Python's True value and not the literal string 'True'.

```suggestion
    projects = Project.objects.filter(is_searchable=True)
```

PR #161, opened 2019-03-14
Add Html Email Framework

marlonkeating:
Up till now we've only sent plain text emails, but for some features going forward we're looking to move to richer, html-based emails.  This change adds a framework for that, as well as changing the email address verification emails to use the framework.

### DIFF
--- a/democracylab/templates/emails/html_email_button.html
+++ b/democracylab/templates/emails/html_email_button.html
@@ -1,4 +1,9 @@
-<!-- BUTTON -->
+<!-- HTML Button template for DemocracyLab emails -->
+<!--
+Template Arguments
+ url: Button link url
+ text: Button text
+ -->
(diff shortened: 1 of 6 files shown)


democracylab/templates/emails/html_email_button.html
```diff
@@ -0,0 +1,25 @@
...
+Template Arguments
+ url: Button link url
+ text: Button text
+ -->
+<!-- Set button background color at TD, link/text color at A and TD, font family ("sans-serif" or "Georgia, serif") at TD. For verification codes add "letter-spacing: 5px;".  -->
+<tr>
+    <td align="center" valign="top" style="border-collapse: collapse; border-spacing: 0; margin: 0; padding: 0; padding-left: 6.25%; padding-right: 6.25%; width: 87.5%;
+			padding-top: 25px;
+			padding-bottom: 5px;" class="button">
+        <a href="https://github.com/konsav/email-templates/" target="_blank" style="text-decoration: underline;">
```
PeterBreen: Need to remove this default link from the template.

PR #162, opened 2019-03-15
change filter key generation to avoid collision error

PeterBreen:
This should fix a bug where if tag display names are not unique across categories, e.g. Software Development: All and Design: All, leading to clicking on one filter activated another instead of the desired filter.

Specifically, filter keys are generated using the guaranteed unique `tag_name` property instead of the `display_name`, which was not guaranteed to be unique. 

To test, ensure that both the above roles are associated with an active project. Attempt to filter by Software Development: All, and make sure that the filter activates in the filter box (is checked) and the filter is added to the querystring. Remove the filter, and repeat for Design: All. 

All other filters should behave normally.


common/components/componentsBySection/FindProjects/Filters/RenderFilterCategory.jsx
```diff
@@ -105,9 +105,9 @@ class RenderFilterCategory<T> extends React.PureComponent<Props<T>, State> {
   _renderFilterList(data) {
     //this function renders individual clickable filter items regardless of category or subcategory status
     let sortedTags = Object.values(data).map((tag) =>
-      <li key={tag.category + '-' + tag.display_name} className="ProjectFilterContainer-list-item">
-          <input type="checkbox" id={tag.category + '-' + tag.display_name} checked={this.props.checkEnabled(tag)} onChange={() => this.props.selectOption(tag)}></input>
-          <label htmlFor={tag.category + '-' + tag.display_name}>
+      <li key={tag.category + '-' + tag.tag_name} className="ProjectFilterContainer-list-item">
```
marlonkeating:
```suggestion
      const key:string = tag.category + '-' + tag.tag_name;
      return  (<li key={key} className="ProjectFilterContainer-list-item">
```

So as to avoid the need to copy-paste any changes in the future.

PeterBreen: Had to do it slightly differently than this exact suggestion but the key is now a const and reused. 👍

PR #164, opened 2019-03-19
Redesign Filter Experience #2

PeterBreen:
Continues work done in #149.

### Changelog:
 - moves "Reset Filters" button above filters, retitled "Clear Filters." Button should always be visible but will be disabled unless at least one filter is active or a search query has been submitted.
 - Moves search box directly above project cards, changes placeholder text.
 - moves location and date modified sorting dropdowns directly above project cards
 - moves and restyles "tag box" below search and sort, restyles selected tag "pills".
 - some technical debt cleanup
 - adds borders and spacing so the whole page should look as design intends
 - adds header text in card section, which changes based on number of active projects after a search is submitted and/or filter is applied ("3 tech-for-good projects found," or similar)

### Does not include:
 - database changes (see [Trello card for details](https://trello.com/c/xl6uNxPJ/161-story-redesign-filter-experience))

### New functionality:
 - "x tech for good projects found..." should say project singular if x = 1. It should be plural if x > 1. x should match the number of shown project cards; if it says "3 projects found" there should be 3 cards.

 - For the 'selected tag' pills, the "x" in a circle in each should be a circle no matter the browser or device OS. Previously different versions of Chrome would render it slightly differently, for example.

### Known Issues:
 - "x tech-for-good projects found" does not work when the location dropdown is used. That is, it won't say "7 tech-for-good projects found" if a user only selects Seattle, or Redmond, etc. Any ideas on this one?


marlonkeating:
Looks like the new text is butting into the list.
image

Also, with the removal of the tag box, the user can't easily see what filters are selected if they navigate to a url with filters.  It would be nice if the filters are expanded in categories with already-selected tags.

common/components/componentsBySection/FindProjects/ProjectSearchBar.jsx
```diff
@@ -26,20 +26,15 @@ class ProjectSearchBar extends React.Component<{||}, State> {
...
     return (
-      <div className="ProjectSearchBar-root col-12">
+      <div className="ProjectSearchBar-root">
         <i className={GlyphStyles.Search}></i>
         <input
           className="ProjectSearchBar-input"
           onChange={e => this.setState({keyword: e.target.value})}
           onKeyPress={this._handleKeyPress.bind(this)}
-          placeholder="Enter keywords"
+          placeholder="Search by issue areas, roles, and more..."
```
marlonkeating: Did this text come from Design?  Because searching in the text search specifically does NOT search issue areas/roles/etc, just project titles and descriptions.

common/components/componentsBySection/FindProjects/ProjectSearchContainer.jsx
```diff
@@ -1,10 +1,10 @@
+//TODO: delete this filf if we choose to remove alerts button. the file is unused right now.
```
marlonkeating: Actually I see that Alerts button has been removed.  Did we make the decision to remove it entirely?

common/components/componentsBySection/FindProjects/ProjectSearchBar.jsx
PeterBreen: We just talked about this in standup but yes, this came from design and I'll change the text to more accurately reflect the search's actual search.

common/components/componentsBySection/FindProjects/ProjectSearchContainer.jsx
PeterBreen: It's been removed from the design but I don't believe we've made a concrete decision one way or the other on deleting it, so I want to leave that todo and file until I get a yes/no.

PeterBreen:
Going back to in-progress until we get a few more items pulled in next sprint.

common/components/componentsBySection/FindProjects/ProjectCardsContainer.jsx
```diff
@@ -22,26 +23,39 @@ class ProjectCardsContainer extends React.Component<{||}, State> {
...
+            {!_.isEmpty(this.state.projects) && <h2 className="ProjectCardContainer-header">{this._renderCardHeaderText()}</h2>}
             {this._renderCards()}
           </div>
         </div>
       </div>
     );
   }
 
+  _renderCardHeaderText(): React$Node {
+    if (this.state.keyword || this.state.tags.size > 0) {
```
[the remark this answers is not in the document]
PeterBreen: I knew I was missing something obvious. That worked, thanks Marlon :)

PR #167, opened 2019-03-21
Add workflow for volunteers to re-engage with projects when they approach projected end date

marlonkeating:
Currently we have volunteers select how long they intend to volunteer with the project, but do nothing with this information.  We want to ask volunteers if they would like to renew their volunteering commitment, or remove them if they choose not to.

Feature Highlights:

- Within 2 weeks of projected end date, volunteers can renew or conclude their volunteer commitment from the My Projects page
- Volunteers will receive 2 emails, a week apart, alerting volunteers their end date is approaching and where to renew
- New header alert also alerting volunteers their end date is approaching
- Metrics for renewing/concluding, and coming to the My Projects page from the reminder emails


civictechprojects/models.py
```diff
@@ -431,6 +443,10 @@ def hydrate_project_volunteer_info(self):
 
     def update_project_timestamp(self):
         self.project.save()
+
+    def is_up_for_renewal(self, now=None):
+        now = now or timezone.now()
+        return (self.projected_end_date - now) < settings.VOLUNTEER_REMINDER_OVERALL_PERIOD
```
PeterBreen: In my dev environment I get an error pointing to this line every time I attempt to log in. I've run the migrations so I'm not sure what it is:

```
TypeError at /index/

unorderable types: datetime.timedelta() < NoneType()

Request Method: 	GET
Request URL: 	http://127.0.0.1:8000/index/?section=FindProjects&showSplash=1
Django Version: 	1.10.7
Exception Type: 	TypeError
Exception Value: 	

unorderable types: datetime.timedelta() < NoneType()

Exception Location: 	/home/dev-admin/git/CivicTechExchange/civictechprojects/models.py in is_up_for_renewal, line 449
```

common/management/commands/volunteer_renewal_reminders.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+
+    email_msg = email_template.render(email_msg, context)
+    send_email(email_msg, settings.EMAIL_VOLUNTEER_ACCT)
+
+
+review_commitment_url = settings.PROTOCOL_DOMAIN + '/index/?section=MyProjects&from=renewal_notification_email'
+
+
+def get_first_email_template():
```
PeterBreen: I mentioned this in standup but I really like this method of making the python readable when HTML email frequently is not. 👍

civictechprojects/views.py
```diff
@@ -355,6 +357,50 @@ def volunteer_with_project(request, project_id):
     return HttpResponse(status=200)
 
 
+# TODO: Pass csrf token in ajax call so we can check for it
+@csrf_exempt
+def renew_volunteering_with_project(request, application_id):
+    if not request.user.is_authenticated():
```
PeterBreen: It's not a big piece of code duplication but if the steps in lines 362-370 and then 386-393 gets more complicated in the future we might want to pull it into its own function to avoid reuse?

civictechprojects/models.py
PeterBreen: Figured it out, my fault. Forgot to set the new env variables.

PR #174, opened 2019-03-27
added border around link icons

ashabrai:


marlonkeating:
For some reason the project Contact/Volunteer buttons now have no space between them.  Looking at the CSS, I'm not sure why that is, but we should fix it.

image

ashabrai:
This change was submitted in another PR

PR #176, opened 2019-03-27
Modify project edits by staff

justBanks:
Refactored to put the user logic in the Project model's save() method (added a user parameter, and made it optional for safety):
```python
def save(self, user=None):
    if user is None or not user.is_staff:
        self.project_date_modified = timezone.now()
    super().save()
```
Also adding a required user parameter in VoluntterRelation.update_project_timestamp():
```python
def update_project_timestamp(self, user):
    self.project.save(user=user)
```
Now, wherever these are called, the logged-in user is passed in


justBanks:
Re-refactor.
Project now has ```update_timestamp()``` function; 
Calling it from the VolunteerRelation view.

civictechprojects/models.py
```diff
@@ -441,13 +445,13 @@ def hydrate_project_volunteer_info(self):
         project_json = self.project.hydrate_to_list_json()
         return merge_dicts(volunteer_json, project_json)
 
-    def update_project_timestamp(self):
-        self.project.save()
+    def update_project_timestamp(self, user):
+        self.project.save(user=user)
```
marlonkeating: This won't actually update the project timestamp anymore now that project_date_modified doesn't update automatically.  Is it even being used anywhere at this point?

justBanks: Agreed. Removed it.

PR #177, opened 2019-03-27
fix navbar

justBanks:
Pertinent lines are 18 and 200 of the changed file


justBanks:
Okay, I think it's good to go

PR #178, opened 2019-03-28
AboutIcon Fits 

ashabrai:


ashabrai:
Marlon, I made the update which you'll see in my last commit. The image should now maintain its ratio whatever size the image would be.

PR #182, opened 2019-04-08
Update to Webpack 4.29.6

PeterBreen:
This branch updates our installed version of webpack to 4.29.6 and rebuilds the site bundle.js with `webpack --mode=production` set, which reduces the size of our bundle.js significantly (down to ~1.18MB)

This branch is currently ready for smoke testing but _should not_ be merged into master without:
 - [x] npm audit
 - [ ] any other package updates 
 - [x] a refactor of webpack's config files - it's implicitly using production mode by default, I want us to have a clear path to build dev or build prod
 - [ ] and finally, full regression testing after the above are completed

stretch goals will be moved to another branch/PR but are preserved here for now:
- [ ] bundle splitting
- [ ] lodash fix
- [ ] tree shaking
- [ ] investigate [webpack-dev-server](https://webpack.js.org/guides/development/#choosing-a-development-tool) so we don't have to manually rebuild for dev changes. Considering our environment we may also need django/webpack integration and react-hot-loader or similar. ([reference](https://owais.lone.pw/blog/webpack-plus-reactjs-and-django/))


PeterBreen:
npm audit returned ~188 issues, most resolved automatically and a couple manually - we're down to two. Both of them are resolvable but have some potential side-effects to evaluate first.

https://www.npmjs.com/advisories/310
https://www.npmjs.com/advisories/786

reference for babel upgrade:
https://babeljs.io/docs/en/v7-migration

reference for node: 
https://davidwalsh.name/nvm

package.json
```diff
@@ -7,10 +7,10 @@
   "author": "Marlon Keating <marlonakeating@gmail.com>",
   "license": "MIT",
   "scripts": {
-    "build": "webpack --config webpack.config.js && npm run build:sass && python manage.py collectstatic --noinput",
+    "build": "webpack --config webpack.config.js && npm run build:sass && python manage.py collectstatic --noinput -i *.scss",
```
PeterBreen: As part of trying to figure out what I want webpack to do and what I want django to do in terms of deployment, I came across [this SO post](https://stackoverflow.com/questions/8269883/how-to-ignore-directories-when-running-django-collectstatic) - `collectstatic` should no longer bring over .scss partials to the production static directory. We can extend that with multiple -i flags if there are other things it brings over that we don't want.

console output after an npm run build should read similar to this:
```
Copying '/home/dev-admin/git/CivicTechExchange/civictechprojects/static/css/compiled/styles.css'
Copying '/home/dev-admin/git/CivicTechExchange/common/static/js/bundle.js'
Copying '/home/dev-admin/git/CivicTechExchange/common/static/js/bundle.js.map'

3 static files copied to '/home/dev-admin/git/CivicTechExchange/staticfiles', 109 unmodified.
```

rather than copying 100+ as it brings over all the scss partials.

package.json
```diff
@@ -7,10 +7,11 @@
   "author": "Marlon Keating <marlonakeating@gmail.com>",
   "license": "MIT",
   "scripts": {
-    "build": "webpack --config webpack.config.js && npm run build:sass && python manage.py collectstatic --noinput -i *.scss",
-    "build:sass": "node-sass civictechprojects/static/css/styles.scss -o civictechprojects/static/css/compiled --output-style compressed",
-    "flow": "flow",
-    "postinstall": "webpack --config webpack.config.js && python manage.py collectstatic --noinput -i *.scss"
+    "build": "webpack --mode development && npm run buildtask:sass && npm run buildtask:collectstatic",
+    "buildprod": "webpack --mode production && npm run buildtask:sass && npm run buildtask:collectstatic",
```
PeterBreen: Until we make some workflow changes in the future (watch/autorebuild, not requiring bundle as part of a branch, etc) the only workflow change is here: When developing, use `npm run build` as normal, but when ready to push a PR use `npm run buildprod` to generate a production-ready bundle.js.

PR #183, opened 2019-04-10
fixed nav bar changes and adjust breakpoint

ashabrai:


civictechprojects/static/css/partials/_SubHeader.scss
```diff
@@ -201,7 +201,7 @@
 // @include media-breakpoint-up(sm) {
 // }
 
-@include media-breakpoint-up(md) {
+@media (min-width: 910px) {
```
marlonkeating: I think we can reduce this amount.  Right before 910 pixels, I can still see a good bit of free space.

image

PR #184, opened 2019-04-10
Upgraded some emails to use HTML instead of Plain Text

harmanbains:
Forgot password
Contact project owner
Volunteer with project
Volunteer leaves project

Have been upgraded to use HTML with the DemocracyLab email template


civictechprojects/views.py
```diff
@@ -495,17 +501,22 @@ def leave_project(request, project_id):
...
+        #     volunteer_name=volunteer_relation.volunteer.full_name(),
+        #     project_name=volunteer_relation.project.project_name,
+        #     message=message)
+        email_template = HtmlEmailTemplate()\
+        .paragraph('{volunteer_name} is leaving {project_name} for the following reason:'.format(
             volunteer_name=volunteer_relation.volunteer.full_name(),
-            project_name=volunteer_relation.project.project_name,
-            message=message)
+            project_name=volunteer_relation.project.project_name))\
+        .paragraph('\"{message}\"'.format(message=message))
```
marlonkeating: The quotes are a nice touch, but now I'm thinking it will stand out even more if the volunteer doesn't enter in any reason for leaving.  I'm thinking we should either say something different when there's no message, or switch to using the conclude email (notify_project_owners_volunteer_concluded_email).

PR #187, opened 2019-04-10
Project Pagination ("More Projects..." button) implemented

TCW417:
Also cleared up React unique key warning I was getting by adding key to React.Fragment in MainHeader.jsx.


democracylab/settings.py
```diff
@@ -169,6 +169,7 @@
 
 USE_TZ = True
 
+PROJECTS_PER_PAGE = 2
```
marlonkeating: Let's make this configurable

common/components/stores/NavigationStore.js
```diff
@@ -10,15 +10,18 @@ import Section from '../enums/Section.js';
 export type NavigationActionType = {
   type: 'SET_SECTION',
   section: SectionType,
+  // page: number,
```
marlonkeating: Are any of the changes on this page needed?

common/components/chrome/MainHeader.jsx
```diff
@@ -53,6 +53,7 @@ class MainHeader extends React.PureComponent<{||}> {
     NavigationDispatcher.dispatch({
       type: 'SET_SECTION',
       section: Section.FindProjects,
+      page: 1,
```
marlonkeating: Why are we setting page on SET_SECTION?

common/components/chrome/SectionLink.jsx
```diff
@@ -6,6 +6,7 @@ import cx from '../utils/cx';
 import NavigationDispatcher from '../stores/NavigationDispatcher.js';
 import React from 'react';
 import url from '../utils/url.js';
+import Section from '../enums/Section.js';
```
marlonkeating: Is this being used?  It looks like we're actually missing the import for SectionType.

common/components/componentsBySection/FindProjects/ProjectCardsContainer.jsx
```diff
@@ -52,6 +60,35 @@ class ProjectCardsContainer extends React.Component<{||}, State> {
             />
         );
   }
+
+  _handleFetchNextPage(e: object): void {
+    e.preventDefault();
+    this.setState({current_page: this.state.current_page + 1 <= this.state.project_pages
```
marlonkeating: This would look cleaner if the current_page logic lived in its own expression.

common/components/componentsBySection/FindProjects/ProjectCardsContainer.jsx
```diff
@@ -52,6 +60,35 @@ class ProjectCardsContainer extends React.Component<{||}, State> {
...
+      ? this.state.current_page + 1 
+      : this.state.current_page }, function () {
+      ProjectSearchDispatcher.dispatch({
+        type: 'SET_PAGE',
+        page: this.state.current_page,
+      });
+    });
+  }
+
+  _renderPagination(): React$Node {
```
marlonkeating:
```suggestion
  _renderPagination(): ?React$Node {
```
To signal that we might return null.

common/components/stores/ProjectSearchStore.js
```diff
@@ -123,8 +139,9 @@ class ProjectSearchStore extends ReduceStore<State> {
   }
 
   _updateFindProjectArgs(state: State): State {
+    let findProjectsArgs: FindProjectsArgs;
     if(state.projectsData && state.projectsData.allTags) {
-      const findProjectsArgs: FindProjectsArgs = _.pickBy({
+      findProjectsArgs = _.pickBy({
```
marlonkeating: Why remove the const and type specification?

common/components/componentsBySection/FindProjects/ProjectCardsContainer.jsx
```diff
@@ -52,6 +60,35 @@ class ProjectCardsContainer extends React.Component<{||}, State> {
...
+    if (this.state.current_page === this.state.project_pages) {
+      return null; // don't render button if we've loaded the last page
+    }
+    return (
+      this.state.projects && this.state.projects.size !== 0
+      ? <div className="page_selection_footer">
+        <button className="page_button" onClick={this._handleFetchNextPage.bind(this)}>
+          More Projects... &rarr;
```
marlonkeating:
```suggestion
          More Projects...
```
Also, we should disable the button while new results are being loaded.

democracylab/settings.py
TCW417: OH! I forgot to reset that to 10.   What do you mean/intend by "configurable"?  Who configures it and how?

common/components/stores/NavigationStore.js
TCW417: Oh, I need to remove those comments.  These are related to removing the page query parameter from the URL

common/components/chrome/MainHeader.jsx
TCW417: Missed that one. I don't think we need that anymore.

common/components/chrome/SectionLink.jsx
TCW417: I haven't messed with this file/these files.  The exports in enums/Section.js look a little strange. My linter goes nuts over the flow syntax so it's hard to tell, but it does indicate that Section isn't used in SectionLink.jsx.

common/components/componentsBySection/FindProjects/ProjectCardsContainer.jsx
TCW417: ok

common/components/stores/ProjectSearchStore.js
TCW417: ? No idea. I don't recall doing that.

common/components/componentsBySection/FindProjects/ProjectCardsContainer.jsx
TCW417: I don't know where that &rarr; came from. I must be inserting random stuff in my keyboard confusion between ubuntu and osx...  
My instance updates so fast I can't see the button until the new projects are loaded so I'm not sure what your concern is re disabling the button while loading...

TCW417:
See updates.

democracylab/settings.py
marlonkeating: By environment variable.  If you check the rest of the settings file, you'll see os.environ.get() lines where we read them in.

common/components/componentsBySection/FindProjects/ProjectCardsContainer.jsx
marlonkeating: Partially to keep people from clicking multiple times and loading multiple pages, and partially to signal to users that something has happened once they've clicked the button and are waiting for the results.  Where I've deployed the changes, there is enough lag where I think we need it.

democracylab/settings.py
TCW417: OK. Got that working. Slacked you an updated .sh file with projects per page included.

common/components/componentsBySection/FindProjects/ProjectCardsContainer.jsx
TCW417: OK. I'll work on it.

TCW417: Current commit on this branch implements waiting for projects to load before enabling the more projects button.  Try it out on your mirror (where presumably you can see some delay) and let me know how it works. I convinced myself via strategic console.logs that it does.

PR #189, opened 2019-04-11
More Projects button behavior (UX) improvements

TCW417:


common/components/componentsBySection/FindProjects/ProjectCardsContainer.jsx
```diff
@@ -94,10 +94,21 @@ class ProjectCardsContainer extends React.Component<{||}, State> {
 
   _renderPagination(): ?React$Node {
     if (this.state.current_page === this.state.project_pages) {
-      return null; // don't render button if we've loaded the last page
+      return (
```
marlonkeating: I don't know that we should show the greyed-out button when there are no more results.  It could give users the mistaken impression that there are more projects waiting to load.

TCW417: Agreed.  I've return it to showing nothing when all projects have been loaded.

TCW417:
Do you want it to show the number of projects currently displayed? because
the # matching search results, with no filters applied, would be the total
number of projects in the system...

marlonkeating:
We want to show how many projects total the current set of filters
returned.  If there are more than 20, we'll want to use the total count
rather than the number that are currently displayed on the page.  We don't
show the count when no filters are selected, so no need to change that.

PR #190, opened 2019-04-12
Tw project count

TCW417:
Added project count value in data returned by server. Updated front end to use it instead of projects.size which is just the number of projects currently displayed.


civictechprojects/views.py
```diff
@@ -206,6 +206,8 @@ def projects_list(request):
         else:
             project_list = projects_by_sortField(project_list, '-project_date_modified')
 
+        project_count = len(project_list)
+        print('project_count', project_count)
```
marlonkeating: Remove count logging

civictechprojects/views.py
```diff
@@ -272,11 +274,13 @@ def projects_by_roles(tags):
     return Project.objects.filter(positions__in=positions)
 
 
-def projects_with_meta_data(projects, project_pages):
+def projects_with_meta_data(projects, project_pages, project_count):
+    print('numProjects', project_count)
```
marlonkeating: Remove logging

democracylab_environment_variables.sh
```diff
@@ -37,7 +37,7 @@ export POSITION_DESCRIPTION_EXAMPLE_URL='https://docs.google.com/document/d/142N
 export VOLUNTEER_CONCLUDE_SURVEY_URL='https://docs.google.com/forms/d/e/1FAIpQLSd4e9FQYX3ARPg7qz1ct5qM_bQW_kkEvQaMuM5LV9Ma1NDgbA/viewform'
 
 # Configure number of projects to display per server request
-export PROJECTS_PER_PAGE=20
+export PROJECTS_PER_PAGE=2
```
marlonkeating: This isn't a good default unless you're specifically testing pagination, so best to keep it at 20 I think.

TCW417: Oops

PR #192, opened 2019-04-17
implement archive models

justBanks:


marlonkeating:
Also, it would be good to not lose all the tags in the Project archive.  If we don't add tag fields to the Project Archive model, then at least we should add a comma-separated string to record what the values were.

PR #194, opened 2019-04-17
Fixes for April 17 2019 Release

PeterBreen:
- changes pagination button behavior to not render at all while projects are loading, changes button theme to reuse existing `btn-theme` style
 - modifies padding rules on project cards to give the subinfo list more room to display, adds some top padding for project description when needed

The padding for subinfo li items was for spacing when they were displayed inline, but now that they're in a more traditional list, it was just making them unnecessarily hard to read. 
Screen Shot 2019-04-17 at 12 31 36 PM
(the green highlight represents the old 1rem padding; this is now 0.2rem to offset the text from the border while giving maximum width possible to the content)


common/components/componentsBySection/FindProjects/ProjectCardsContainer.jsx
```diff
@@ -95,20 +95,13 @@ class ProjectCardsContainer extends React.Component<{||}, State> {
   }
 
   _renderPagination(): ?React$Node {
-    if (this.state.current_page === this.state.project_pages) {
+    if ((this.state.current_page === this.state.project_pages) || this.state.projects_loading) {
```
marlonkeating: I think it's still useful to show Loading... after the user clicks on the button.

PeterBreen: That makes sense. It's hard for me to test that locally because my data loads basically instantly but I wrote a conditional that should handle it. My thought was if this.state.projects isn't empty (so it's not showing the button on initial pageload) and the loading state is true, it should show the "Loading..." button in a disabled state.

When I manually set the state this way in React it worked, but it needs testing with live data.

common/components/componentsBySection/FindProjects/ProjectCardsContainer.jsx
```diff
@@ -95,9 +95,18 @@ class ProjectCardsContainer extends React.Component<{||}, State> {
   }
 
   _renderPagination(): ?React$Node {
-    if ((this.state.current_page === this.state.project_pages) || this.state.projects_loading) {
+    if (this.state.current_page === this.state.project_pages) {
```
marlonkeating: nit: When loading the last page of results, the button disappears

PeterBreen: It should render the Loading button even on the last page of results now, so long as this.state.projects_loading is true.

PR #195, opened 2019-04-24
added background color to the volunteer application buttons

ashabrai:


civictechprojects/static/css/partials/_Global.scss
```diff
@@ -18,6 +18,10 @@ body {
 a.btn-theme:hover, a[type="button"].btn-theme:hover {
   text-decoration: none;
 }
+.btn {
```
marlonkeating: I would add the btn-theme class to the specific buttons, rather than altering a style that's used by just about every button on the site.

marlonkeating:
Closing since button designs have changed.

PR #202, opened 2019-05-01
Implement archiving using custom model Manager

justBanks:


justBanks:
Just occurred to me... No need to save the record twice

civictechprojects/models.py
```diff
@@ -34,7 +34,30 @@ class TaggedOrganization(TaggedItemBase):
...
+class ArchiveManager(models.Manager):
+    def get_queryset(self):
+        return super(ArchiveManager, self).get_queryset().filter(deleted=True)
+
+
+class DefaultManager(models.Manager):
+    def get_queryset(self):
+        return super(DefaultManager, self).get_queryset().filter(deleted=False)
+
+
```
marlonkeating:
```suggestion
# This base class adds delete functionality to models using a flag,  and filters deleted items out of the default result set 
```

PR #207, opened 2019-05-08
update navbar

justBanks:


common/components/controllers/CreateProjectController.jsx
```diff
@@ -35,13 +36,14 @@ class CreateProjectController extends React.PureComponent<{||},State> {
   }
   
   render(): React$Node {
-    return (
+    return !CurrentUser.isLoggedIn() ?
```
marlonkeating: Look at what we do for the Create Project button when users aren't logged in.  We link them to the Login page, which then redirects them to the Create Project page once they've logged in.  With your solution, the user is redirected to the Find Projects page once they've logged in, not the Create Project page.

common/components/chrome/MainHeader.jsx
```diff
@@ -143,6 +143,18 @@ class MainHeader extends React.Component<{||}, State > {
...
   }
 
+  _renderLogInLink(): void {
+    return (
+      <SectionLink
+          activeSection={this.state.activeSection}
+          key="LogIn"
+          section={Section.LogIn}
+          title="Log In"
+          target="_blank" rel="noopener noreferrer"
```
marlonkeating: This line doesn't do anything since it's not a proper link, and If it did, the Login page would open in a new tab which we don't want.

justBanks:
Alright, I'm giving it another go with the latest commit...

common/components/controllers/CreateProjectController.jsx
justBanks: Alright, I'm giving it another go with the latest commit...

common/components/chrome/MainHeader.jsx
```diff
@@ -113,13 +113,13 @@ class MainHeader extends React.Component<{||}, State > {
           <div className={this._cx.get('rightContent')}>
             {this._renderSectionLinks()}
             {this._renderHeaderLinks()}
+            {this._renderHeaderButtons()}
+            <div className="vertical-divider"></div>
             {
               CurrentUser.isLoggedIn() ?
                 this._renderAccountInfo() :
-                this._renderLogInButton()
```
marlonkeating: Remove the functions if we're not using them anymore.

justBanks: Addressed everything in the latest commit

PR #208, opened 2019-05-10
Peter/aboutpage3

PeterBreen:
Refining the about page after #203 and #204:

Task tracking: https://trello.com/c/Ii3qimcb/247-about-us-page-additions


common/components/controllers/AboutUsController.jsx
```diff
@@ -7,11 +7,13 @@ import type {ProjectDetailsAPIData} from '../utils/ProjectAPIUtils.js';
...
 import Headers from "../common/Headers.jsx";
 import Person from '@material-ui/icons/Person';
-// TODO: Create metrics event for this page, then import metrics and use it
-// import metrics from "../utils/metrics.js";
+import BioModal from "../componentsBySection/AboutUs/BioModal.jsx";
 
 type State = {|
   aboutUs: ?ProjectDetailsAPIData,
+  showBiographyModal: boolean,
+  modalPerson: object,
```
marlonkeating: If modalPerson accepts multiple types, you can use Flow unions.  https://flow.org/en/docs/types/unions/

PeterBreen:
> I think we need to cap the dimensions of the bio modal and probably put a scroll bar on long biographies. If you open the About page in a mobile browser and click on someone with a long bio, it will fill up the screen and be very hard to click out of the modal.

Agreed, it's awkward. I can add another close button to the bottom of the modal to help people escape, but I'm also seeing some unusual scrolling behavior; possibly due to the `top: 22%` rule we apply to all the modals sitewide which I think was put in place so they don't get cut off at the top of the browser window, because ... I don't know why but they do.

Bootstrap's [browser and device support](https://getbootstrap.com/docs/4.0/getting-started/browsers-devices/#modals-and-dropdowns-on-mobile) page lists a few issues with modals on mobile that mean no matter what I do we'll have some problems, but I will fix what I can.

common/components/controllers/AboutUsController.jsx
PeterBreen: I read the flow docs you linked on unions and I think I've set it correctly. All the key/value pairs inside the modalPerson object are numbers or strings. I'm not sure though, so if I got that wrong just let me know.

PeterBreen:
Updated to address both issues. The modals are going to be hard; what I'd like to try at this point is vertically-centered modals but our react-bootstrap doesn't appear to support them, we'll need to upgrade. I was not successful so far in figuring out a way to position them differently than we already do, so the only substantive change to the modals should be the close button at the bottom.

common/components/componentsBySection/AboutUs/BioModal.jsx
```diff
@@ -14,6 +15,9 @@ type State = {|
   showModal: boolean,
 |};
 
+
+const defaultBiography = "This user hasn't provided us with a biography yet. But trust me, when they do it'll be awesome.";
```
marlonkeating: Has this text been vetted yet?

PeterBreen: haha not in the slightest, no. That was my test string for the code, it shouldn't make it to master, good catch.

PeterBreen:
> Having the Close button down below on the modal makes it closable on mobile, but it can be hard to effectively see the person's name, and click on the link to their profile.

I think I *finally* figured this one out. 

Here's a snippet of code from bootstrap's modal partial:
```css
  // When fading in the modal, animate it to slide down
  .modal.fade & {
    @include transition($modal-transition);
    transform: translate(0, -25%);
  }
  .modal.show & {
    transform: translate(0, 0);
  }
```

The problem? Just like before, the fade class _never leaves_ in our current implementation, meaning that translate(0, -25%) kept making our modals position strangely.

I fixed it by manually adding the `show` class to my modal and then removing the `.modal-dialog { top: 22%}` rule I mentioned earlier - this is a hack. In another PR, I intend to override this bootstrap code entirely so that it will be fixed for all modals on the site, but I wanted my fix to be self-contained to the about me page for now.

PR #215, opened 2019-05-16
realign navbar items

justBanks:


justBanks:
I think the latest commit should look better

PR #218, opened 2019-05-22
re-route MyProjects to LogIn page

justBanks:


common/components/controllers/LogInController.jsx
```diff
@@ -13,7 +13,8 @@ type Props = {|
 type State = {|
   username: string,
   password: string,
-  errorMessage: string
+  prevPage: string,
+  //errorMessage: string
```
marlonkeating: Go ahead and delete variables we aren't using.

PR #222, opened 2019-05-29
Add Project approval workflow

marlonkeating:
Projects will now require admin approval going forward.  This change does the following:
- Newly-created projects are not visible on the site until approval, except for admins and their creator
- Admins receive a notification email when a new project is created, which has a button that when clicked will approve the project
- Project creators receive email notification when their project is approved


civictechprojects/forms.py
```diff
@@ -80,6 +81,8 @@ def create_project(request):
         # Notify the admins that a new project has been created
         send_project_creation_notification(project)
 
+        messages.add_message(request, messages.INFO, 'Your project is awaiting approval.  Expect a decision in the next 24 hours.')
```
PeterBreen: In case someone has multiple projects on the site can we get the project name in this message? 

also just for user expectations you might want to say 'next business day' so approval admins aren't theoretically obligated to approve things on a Saturday night. Even if people end up doing that it's a lot easier to underpromise and overdeliver.

PR #231, opened 2019-06-11
Peter/vol ui fix

PeterBreen:
handles requested Volunteer Management UI Improvements: https://trello.com/c/JZxCfnwU/264-volunteer-management-ui-improvements

 - adds icons to dropdown links 
 - moves pending volunteers above the rest of the team if applicable (they exist and the logged in user has admin or co-owner status)

The way it handles that second point is pretty messy; it needs a more thorough refactor but if there are any quick optimizations to be made, let me know so I can get them in before code freeze for this sprint.


common/components/common/volunteers/VolunteerCard.jsx
```diff
@@ -62,21 +66,21 @@ class VolunteerCard extends React.PureComponent<Props> {
   _renderApplicationMenuLinks(): ?Array<React$Node>  {
     if (this.props.volunteer && this.props.volunteer.isCoOwner) {
       return [
-          (<MenuItem className="VolunteerCard-caution" onSelect={() => this.props.onDemotionButton(this.props.volunteer)} key="0">Demote</MenuItem>),
-          (<MenuItem className="VolunteerCard-danger" onSelect={() => this.props.onDismissButton(this.props.volunteer)} key="1">Remove</MenuItem>)
+          (<MenuItem onSelect={() => this.props.onDemotionButton(this.props.volunteer)} key="0"><i className={`${GlyphStyles.Pushpin} fa-fw`}></i> Demote</MenuItem>),
```
marlonkeating: We should add functionality for adding 'fa-fw' to glyph classes, similar to how we accommodate glyph sizes.  Add a TODO if you don't feel like doing it this time around.

common/components/utils/glyphs.js
```diff
@@ -43,4 +43,26 @@ export function Glyph(style: string, size: ?string): string {
...
+  }
+  let defaults = {
+    size: "",
+    fixedWidth: false
+  };
+  options = setDefaults(options, defaults);
+
+  //now modify style based on passed or default options
+  if (options.size) {
+      style = style += (' ' + options.size)
```
marlonkeating: Isn't style += enough?

common/components/utils/glyphs.js
```diff
@@ -43,4 +43,26 @@ export function Glyph(style: string, size: ?string): string {
   return style + (size ? " " + size : "");
 }
 
+//to use GlyphOption you must also import GlyphStyles, GlyphSizes, from glyphs.js
+export function GlyphOption(style: string, options: object) {
```
marlonkeating: This solution works.  There's a bit more code than I'd prefer.  If I had to make it work with a minimum of code, what I would do is change the Glyph function to accept a variable number of style add-ons, like
export function Glyph(style: string, ...args)

So we can just tack on extra style parameters as needed.  Ex:
Glyph(GlyphStyles.Pushpin, GlyphSizes.X2, GlyphWidth.Fixed)

With the added bonus that we wouldn't have to change any existing calls to Glyph()

common/components/utils/glyphs.js
PeterBreen: It is, yeah. I'll fix that.

common/components/utils/glyphs.js
PeterBreen: I just pushed something that should do this. Array.join didn't take to me using " " as a separator (empty strings are ignored) so I just modified GlyphSizes and GlyphWidth to include the spaces we need for the class declarations.

PR #232, opened 2019-06-11
Sort project links by type

marlonkeating:
Currently links are displayed in the order they were entered, but with this change we will highlight the special links (code, messaging, project management, file repository) above the rest.


common/helpers/collections.py
```diff
@@ -4,3 +4,4 @@ def find_first(collection, condition):
         if condition(item):
             return item
     return None
+
```
marlonkeating:
```suggestion

```

PR #233, opened 2019-06-13
Fixes for 2019-06-13 release

PeterBreen:
Fixes for release: 
 - fixes two React warnings on any About Project page
 - fixes site navigation display error between 925-990px display width
 - style fixes for About Project page on mobile / narrow viewport devices


civictechprojects/static/css/partials/_SubHeader.scss
```diff
@@ -199,7 +199,7 @@
   margin-left: 25px;
 }
 
-@media (min-width: 925px) {
+@media (min-width: 990px) {
```
marlonkeating: With the new font I think we need to break at 1000px

image

PeterBreen: Which browser/OS is this, out of curiosity? I tested a few locally and found 985 was the widest breakpoint so I went to 990 for a buffer, clearly I needed to test a few more cases!

marlonkeating: Chrome/Win10

PR #242, opened 2019-06-26
Peter/social footer

PeterBreen:
- Adds a new footer to all pages on the site, per this: https://trello.com/c/UNRkFHYm/251-social-newsletter-footer
 - Adds MailChimp newsletter signup form in new component for potential reuse
 - Removes Contact Us from top nav and places it here instead
 - three breakpoint design; single column (narrow devices, like phones), two column (tablets) and three column (desktop)

Known Issues:
 - [x] background image or colors from Design
 -  [x] content color styles including normal/hover states for buttons, social media icons (will be done after the first is resolved)
 - [x] Subscribe button isn't inline with the input field for email address


PeterBreen:
Ready for review. This branch requires #247 for the Contact link to work correctly.

common/components/common/integrations/NewsletterSignup.jsx
```diff
@@ -0,0 +1,34 @@
+// @flow
+
+import React from 'react';
+
+class NewsletterSignup extends React.Component {
+
+render() {
+  return (
+    <div id="mc_embed_signup">
+      <form action="https://democracylab.us3.list-manage.com/subscribe/post?u=72af92d0a817dcbf3aa960ee0&amp;id=d3b4c4d81c" method="post" id="mc-embedded-subscribe-form" name="mc-embedded-subscribe-form" className="validate" target="_blank" noValidate>
```
marlonkeating: Are there any mailchimp id's in here that would make sense to put in an environment variable?

common/components/common/integrations/NewsletterSignup.jsx
```diff
@@ -0,0 +1,34 @@
+// @flow
+
+import React from 'react';
+
+class NewsletterSignup extends React.Component {
+
+render() {
+  return (
+    <div id="mc_embed_signup">
```
marlonkeating: Fix indentation in this html.

common/components/common/integrations/NewsletterSignup.jsx
PeterBreen: Right now, I don't think so, but possibly in the future. I'm not entirely familiar with what Mailchimp needs to be unique per instance of a signup form, so I'd rather not set variables that may not need to be variables until we have a need.

PeterBreen:
> Looks good overall. For mobile, I think we might want to put the Subscribe button below the box.
> 
> image

Okay, the latest commit should fix this. Here's how it ought to work:
Mobile (1 column): Button is centered below the input email field
Tablet (2 column): Button is left-aligned below the input email field
Desktop (3 column): Button is inline with the input field, with space between.

PR #243, opened 2019-06-28
Google conversion tracking

marlonkeating:
- Add google ads snippet to all pages
- Add google ads conversion snippet for page that appears after user signs up for the site


common/helpers/front_end.py
```diff
@@ -1,7 +1,15 @@
+import re
```
PeterBreen: Is this linebreak intentional?

PR #248, opened 2019-07-16
Unit tests

justBanks:
This update requires running `pip install -r requirements.txt`


justBanks:
fixed that for you

justBanks:
The existing `Glyph` function works fine when just one additional argument is passed. 
But, if an additional style is specified, the `join()` function will insert a comma. Changing that to `join('')`

PR #249, opened 2019-07-17
added lodash _.throttle function in EditProjectForm.jsx line 373 to p…

Lennerblom:
…revent fast multi click when submitting new project.this prevents duplicate new project.


common/components/common/projects/EditProjectForm.jsx
```diff
@@ -370,7 +370,7 @@ class EditProjectForm extends React.PureComponent<Props,State> {
         <div className="form-group pull-right">
           <div className='text-right'>
             <input disabled={!this.state.formIsValid} type="submit" className="btn_outline save_btn"
-                   value="Save Project" onClick={this.onSubmit.bind(this)}/>
+                   value="Save Project" onClick={_.throttle(this.onSubmit.bind(this), 1000,{'trailing': false})}/>
```
PeterBreen: Can you explain why you chose throttle over debounce here? debounce seems better to me, based on [this article](https://css-tricks.com/debouncing-throttling-explained-examples/), but I might be missing something.

Lennerblom: I tried to use debounce first, but I wasn't able to make it work.  I actually read that same article in your link and then tried throttle...then it worked.  I'll admit I don't know why the debounce didn't work...

PeterBreen: I've gotten debounce to work before in a react app where instead of binding this inline, like here, it was in the constructor, something like this:
```
 constructor(props) {
     super(props);
     this.onSubmit = _.debounce(this.onSubmit.bind(this),1000, { 'leading': true });
...
```
and then on line 373 you'd just have `onClick={this.onSubmit}`

I think react's event pooling is why this current implementation isn't working out for you but i'm not expert on how that works so take it with a grain of salt :)

PeterBreen:
Closing due to being superseded by #259

PR #250, opened 2019-07-18
Create project revamp

marlonkeating:
Updating the current create project page to use a multi-page workflow laid out by the design team.

Pending Items:
[X] Fix tag selector issues
[X] Put project id in url after first page so reloads will bring user to project being created
[X] Disable 'Edit' buttons on About Project preview
[X] Show notification that project will be approved within 24 hrs
[X] Metrics
[X] Code/logging cleanup


civictechprojects/static/css/partials/_CreateProjectForm.scss
```diff
@@ -1,4 +1,12 @@
-//overrides a very particular user agent stylesheet declaration
+.form-body {
```
PeterBreen: Not blocking, but it might be worth making the classname less generic than form-body. create-form-body? Or sidestep it entirely; targeting `.create-form h1` and `.create-form h2` would probably do fine as well - whatever makes the most sense to you.

marlonkeating: In this case I think generic is best because the design team's mockups for other planned creation pages have the same header layout.

PR #253, opened 2019-07-23
Added project solution and actions description fields

wenyi189:
https://trello.com/c/a4h94lzz/299-add-project-solution-actions-description-fields


common/components/common/projects/AboutProjectDisplay.jsx
```diff
@@ -222,7 +222,11 @@ class AboutProjectDisplay extends React.PureComponent<Props, State> {
 
           <div className='AboutProjects-details'>
             <div id='project-details'>
-              {project.project_description}
+              {
+                project.project_description + "\n" +
```
marlonkeating: I think it makes more sense to have a double space between paragraphs.  Our existing projects tend to use double spaces when they're typing up their descriptions.

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -318,13 +324,37 @@ class EditProjectForm extends React.PureComponent<Props,State> {
             }
           </label>
           <div className="character-count">
-            { (this.state.formFields.project_description || "").length} / 3000
+            { (this.state.formFields.project_description || "").length} / 1000
```
marlonkeating: For this field we're going to want to keep the existing 3000 character limit in order to accommodate projects who already have long descriptions.

common/components/common/projects/AboutProjectDisplay.jsx
```diff
@@ -222,11 +222,13 @@ class AboutProjectDisplay extends React.PureComponent<Props, State> {
 
           <div className='AboutProjects-details'>
             <div id='project-details'>
-              {
-                project.project_description + "\n" +
-                project.project_description_solution + "\n" +
-                project.project_description_actions
-              }
+              {project.project_description}
+              <br></br>
```
marlonkeating: Actually, since Solution and Actions are optional fields, we should check to see that they are present before adding the padding between them.  Otherwise, we could end up putting in a bunch of blank space we don't need.

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -324,11 +325,13 @@ class EditProjectForm extends React.PureComponent<Props,State> {
             }
           </label>
           <div className="character-count">
```
marlonkeating: Let's put the character counts below the fields and to the right per the design.

PR #255, opened 2019-07-24
7/23/19 bugfixes

marlonkeating:
- Clicking 'Back' to go to My Projects page was failing due to an error where a flux dispatch was being called while another flux dispatch was ongoing
- Save button on Edit Project and Edit Profile was cutting into the new footer


common/components/controllers/EditProfileController.jsx
```diff
@@ -160,7 +160,7 @@ class EditProfileController extends React.PureComponent<{||},State> {
       <div className="wrapper-gray">
         <div className="container">
           <form action={`/api/user/edit/${window.DLAB_GLOBAL_CONTEXT.userID}/`} method="post">
-            <div className="EditProjectForm-root">
+            <div className="EditProjectForm-root create-form white-bg container-fluid">
```
PeterBreen: I don't think we need the `container-fluid` class here. Line 161 already has this entire section in a `.container` so we're good already.

civictechprojects/static/css/partials/_EditProjectForm.scss
```diff
@@ -2,3 +2,8 @@
 .btn-background-project, .btn-background-project input[type="button" i] {
     background-color: $color-grey-medium;
 }
+
+.EditProjectForm-root .save_btn {
```
PeterBreen: This works to fix the symptom but I think we can address the root issue (a floated element floating outside its container) without it. Good news: We can just delete the float and fix the problem, since we align this button to the right twice. There's both a `pull-right` (float: right) and `text-right` (text-align: right) class applied to the button, so just remove pull-right and there's no need to relatively reposition because there's no more float.

Remove this CSS declaration entirely and then change EditProjectForm.jsx (line 369) and EditProfileController.jsx (line 245) from this:
`<div className="form-group pull-right">` 
to this:
`<div className="form-group">`

That fixes it without having to put more style rules on top.

common/components/common/projects/EditProjectForm.jsx
```diff
@@ -221,7 +221,7 @@ class EditProjectForm extends React.PureComponent<Props,State> {
 
   _renderForm(): React$Node {
     return (
-      <div className="EditProjectForm-root">
+      <div className="EditProjectForm-root create-form white-bg container-fluid">
```
PeterBreen: We don't need this `.container-fluid`declaration because EditProjectController already encases this entire component in a `.container` class.

civictechprojects/static/css/partials/_EditProjectForm.scss
marlonkeating: Thanks, your suggestion worked like a charm!

PR #256, opened 2019-07-24
Added Project Saved alert

kschecht:
When the user clicks the "Next" button, the text "Project Saved" with a checkmark next to it will appear. This text will disappear if the user clicks the "Back" button or edits a text field. (In the future, it should disappear if a non-text field is edited.) The color scheme of the "Next" button has also been modified to make it orange, like the other buttons and colors on the page.


common/components/controllers/CreateProjectController.jsx
```diff
@@ -177,7 +187,16 @@ class CreateProjectController extends React.PureComponent<{||},State> {
   }
   
   onFormUpdate(formFields: {||}) {
-    this.setState({fieldsUpdated: true});
+    if (!this.state.clickedNext && !this.compareFormFields(this.state.currentFormFields, formFields)) {
```
marlonkeating:
```suggestion
    if (!this.state.clickedNext && !_.isEqual(this.state.currentFormFields, formFields)) {
```
Did a little research, and it looks like _.isEqual will do the trick in lieu of our custom function.

PR #258, opened 2019-07-30
add Mission statement to bottom of landing image

PeterBreen:
- adds mission statement text
- positions at the bottom of the splash screen image
- constrains maximum width so it should stay as two lines of text even on high-res displays


PeterBreen:
Just pushed a commit which redoes how the content goes in the splash. Works for me in my test but it'll require test at all viewport sizes.

Also removed the old 501(3)(c) text that was between the H1 and the buttons, changed the Find Projects button since Mark mentioned it and I was in this file anyway.

PR #260, opened 2019-08-01
Project owners can send messages to volunteers

marlonkeating:
- Project owners can email all volunteers by clicking 'Contact Volunteers'
- Project owners can email single volunteers by clicking 'Contact' within the volunteer context menu


PeterBreen:
Three changes I'd like to see, none critical:
 - If there's a misconfigured set of variables on the back end (source: me, and my misconfigured variables, oops) it hangs endlessly on the front at the sending message state - no failure is reported back to the frontend. Ideally if it returns anything other than 200 OK we could close the modal and return an error message, but that might be a larger item for another sprint.

These two are probably quick:

 - Contact Volunteers button shows up even for projects with 0 volunteers. Might be confusing to a project owner?

 - "Contact Volunteers" button has margin-right 13px, should be 8px to align with the button above it.

PR #282, opened 2019-08-17
Topic image dialog upgrade - drag and drop file selector and image cropper

mattiejean:
Libs added: react-file-drop and react-image-crop

FileSelectButton allows file selection with a button or drag and drop. It also has optional parameters hasImagePreview and previewImage, which, if set lets the parent component set a background image on the drop area.

 ImageCropUploadButton - has a FileSelectButton and a ReactCrop area.  Once an image is selected its sent to the cropper.  Once the user selects their crop and hits the done button the image is uploaded to S3.

ImageCropUploadFormElement is just a copy of ImageUploadFormElement but with an ImageCropUpload button.


common/components/common/upload/ImageCropUploadButton.jsx
```diff
@@ -0,0 +1,197 @@
...
+  constructor(): void {
+    super();
+    this.state = {
+      s3Key: "",
+      src: null,
+      isCropping: false,
+      crop: {
+        unit: "%",
+        width: 30,
+        aspect: 16 / 9
```
marlonkeating: Aspect should be a prop, as not all images will have the same ratio.  The Project icon should be 1:1, for instance.

mattiejean:
I pushed up everything except the white text for "Drag your images here".  I wasn't sure exactly what you wanted here. Do you want white, bold text with another color background so no matter what color the image is behind it it shows up?

common/components/forms/ImageCropUploadFormElement.jsx
```diff
@@ -45,9 +45,10 @@ class ImageCropUploadFormElement extends React.PureComponent<Props,State> {
   }
 
   render(): React$Node {
+    const aspect = 16 / 9;
```
marlonkeating: ImageCropUploadFormElement should also have a prop for aspect ratio, defaulting to 1:1 for now since that is the image format our existing cases (project and user thumbnail) use.

civictechprojects/static/css/partials/_ImageCropUploadButton.scss
```diff
@@ -0,0 +1,18 @@
+.ImageCropUploadButton-cropper {
+    position:relative;
+    border: 1px solid black;
+    border-radius:10px;
+    /* width:500px;
```
marlonkeating: Can delete these lines if we're not going to use them.

marlonkeating:
For the Drag your images here text, in order to make it legible with all images, what we could use is white text over a transparent black overlay.

common/components/common/upload/ImageCropUploadButton.jsx
```diff
@@ -0,0 +1,201 @@
...
+  _handleDoneCropping() : void {
+     this.setState( { isCropping: false } );
+     if (this.fileBlob) {
+        this.launchPresignedUploadToS3(this.fileBlob);
+     }
+  }
+
+  _onCropChange(crop, percentCrop) : void {
+    // You could also use percentCrop:
+    // this.setState({ crop: percentCrop });
```
marlonkeating: Do we need this?

common/components/common/upload/ImageCropUploadButton.jsx
```diff
@@ -0,0 +1,201 @@
...
+      0,
+      0,
+      crop.width,
+      crop.height
+    );
+
+    return new Promise((resolve, reject) => {
+      canvas.toBlob(blob => {
+        if (!blob) {
+          //reject(new Error('Canvas is empty'));
```
marlonkeating: Do we need this?

common/components/common/upload/FileSelectButton.jsx
```diff
@@ -0,0 +1,79 @@
...
+      reader.addEventListener('load', () => {
+        this.setState({ src: reader.result });
+        this.props.onFileSelect(reader.result);
+      });
+      reader.readAsDataURL(file);
+    
+  };
+
+  _isValidFileType(file) : boolean {
+    // TODO check against this.props.acceptedFileTypes
```
PeterBreen: Are we okay with this waiting for a future pr?

common/components/common/upload/ImageCropUploadButton.jsx
```diff
@@ -0,0 +1,188 @@
...
+            onChange={this._onCropChange.bind(this)}
+          />
+            <input type="button" value={this.props.cropButtonText || "Done Cropping"} onClick={this._handleDoneCropping.bind(this)} className="ImageCropUploadButton-doneButton"/>
+</div>
+        )}
+        {!this.state.isCropping && (
+          <FileSelectButton hasImagePreview="true" 
+             previewImage={previewImage}
+             acceptedFileTypes="image/*"
+             dragText={this.props.dragText || "Drag Your Images Here or"}
```
PeterBreen: Do we support multiple upload for images? Maybe this should read image singular?

marlonkeating:
```suggestion
             dragText={this.props.dragText || "Drag Your Image Here or"}
```

common/components/common/upload/FileSelectButton.jsx
marlonkeating: I think that's fine.  The default behavior of the react library seems to default to not letting users select non-image files.

marlonkeating:
Approving for now, will address UI polish issues later: https://trello.com/c/IVfYXWY5

PR #284, opened 2019-09-18
Peter/filter flash

PeterBreen:
trello item: https://trello.com/c/wQ7mhDRU/316-find-projects-clicking-filter-makes-other-selected-filters-flash-white-for-a-second-while-new-search-results-load

From what I've been able to determine the filter components, due to how they held state, were re-rendering from scratch every time something changed, causing the blink. Specifically, because the state object which held checked/unchecked data didn't know or care about the categories of filter, neither did the rerender; some console.logs demonstrated the isChecked function would fire on every filter item if any of them changed; that's a lot of needless re-render.

To mitigate this, I moved state-control for the individual filter items into the child `<RenderFilterCategory />` component instead of the parent `<ProjectFilterDataContainer />` component. Each child component handles pushing to and from the flux store's selectedTags object (which is our 'source of truth') 

Or that's the intent. There are two outstanding issues, both of which I think are caused by me not using that source of truth properly. Clicking a filter will add and clicking it again will remove, but the Clear Filters button and clicking the X on each selected filter pill does not work. 

On the upside, in my testing the filter flashing is totally gone. And I was just expecting to minimize it, not remove it. so if I can get this to listen to changes from the Reset Search button and clicking the X inside a filter pill...

### DIFF
--- a/common/components/componentsBySection/FindProjects/Filters/ProjectFilterDataContainer.jsx
+++ b/common/components/componentsBySection/FindProjects/Filters/ProjectFilterDataContainer.jsx
@@ -57,7 +57,6 @@ class ProjectFilterDataContainer extends React.Component<Props, State> {
         sortedTags: sorted
       });
     });
-    this._selectOption = this._selectOption.bind(this);
   }
 
   static getStores(): $ReadOnlyArray<FluxReduceStore> {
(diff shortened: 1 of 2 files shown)


PeterBreen:
Closing this due to tackling the issue in another way; the implicit filter checked/unchecked state of using a function to "look at a selectedTags object and see if I'm in there" rather than a specific key with a true/false value is a block I'm unable to resolve without refactoring further up the chain (most likely in the flux store directly)

PR #286, opened 2019-09-21
Peter/check enabled fix

PeterBreen:
For testing whether or not this native js check is faster than lodash _.has()


common/components/componentsBySection/FindProjects/Filters/ProjectFilterDataContainer.jsx
```diff
@@ -122,8 +122,11 @@ class ProjectFilterDataContainer extends React.Component<Props, State> {
     }
 
     _checkEnabled(tag: TagDefinition): boolean {
-      //return true if tag is in this.state.selectedTags, else implicitly false
-      return _.has(this.state.selectedTags, tag.tag_name)
+      if (this.state.selectedTags[tag.tag_name]) {
```
marlonkeating:
```suggestion
      return !!this.state.selectedTags[tag.tag_name];
```

PR #295, opened 2019-11-06
Peter/reactstrap4

PeterBreen:
Upgrades react-bootstrap to target v4, resolving several conflicts we had with react-bootstrap looking for v3 and our served CSS being v4. Themes Bootstrap to use our colors, changes text color on DL orange background for readability, fixes selected component layout and templating issues as related to all of the above.

Specifically, these upgrades (usually via yarn upgrade --latest)
 - react-bootstrap v1.0.0-beta.14 (the latest) which targets Bootstrap 4.3
 - Bootstrap CSS to v4.3.1 (also latest) 
 - React 16.11 (latest) because we needed to be >16.8 for react-bootstrap
 - React DOM 16.11 (latest) as above

After auditing the site, I found we use four react-bootstrap components, `<Button>`, `<Modal>`, `<Form>`, and `<Dropdown>`. I've gone through and rewritten every one I could find to conform to the new patterns (when required, modals saw no changes for example) 

All elements of our site which use or imitate Bootstrap components will need to be tested to make sure we're assigning the proper variants, using the right props for size/positioning/etc, and that all states operate the way we want: disabled, selected, active, hover, and so on. Of note is that we are replacing a lot of pre-defined Bootstrap default colors with our own; btn-primary is no longer default Bootstrap blue but DemocracyLab orange. These keys are used by Bootstrap in a variety of ways so we need to check a lot of items. 

Or, in short, "it might look a little weird."

## known issues (bugs to fix)

Anything marked with (OLD) is to the best of my knowledge not due to this PR but still worth fixing. They will be prioritized last for this week's release and are not blocking if unfixed, but I'll try to get them anyway.

#### Find Projects

- [x] clear filters button has a theme orange border  in disabled state (check other disabled state buttons?)
- [x] selected filter 'pills' have white text on orange, change to black
- [x] selected filters in filter component have white text on orange, change to black
- [x] (OLD) change .page_selection_footer { margin-top: 15px; } to {margin: 15px 0; }

#### Project Profile
- [x] Volunteer with project modal has a nested dropdown for the "Position To Apply For" field
- [x] Dropdown menu with administrative for a volunteer aren't styled right (always orange, no color-coding)
- [x] clicking the "X" in the role selector triggers an `Uncaught TypeError: Cannot read property '0' of null`
- [x] (OLD) Jumplink for Skills Needed goes too far down the page

#### Create Project
- [x] Next button doesn't change look on hover/isn't styled with btn-primary
- [x] Project Links, Project Files and Roles Needed buttons aren't styled correctly (should not be grey if active)
- [x]  create project "back" and "next" buttons aren't aligned in the same row properly
- [ ] (OLD) the edit/remove icons for a link, file, or role needed don't change the cursor or react to hover, are too tightly spaced
- [ ] (OLD) the edit/remove icons for a link, file, or role needed should have inline text (e.g. "Edit" or "Remove") instead of just icon (accessibility)

#### Donate/Thank You page
- [x]  donate page buttons are all wrong (should be outline black text, outline orange on hover)
- [ ] (OLD) change Donate with PayPal button from image to button (if possible) or update style if not
- [ ] (OLD) Refactor pages to use bootstrap container classes

#### User Profile 
- [x] nested dropdown on "Country" field of user profile
- [x] Resume/Links/File butons have the same issue as Create Project icons
- [x] update "Save Changes" button to btn btn-success class, remove custom styling
- [x]  user profile page links are blue (not orange) and need padding at the bottom of the section
- [x]  user profile page has inline margin values that should be replaced with col-xs-12 col-md-? classes

#### Contact Us
 - [x] send button has incorrect border when in disabled state

#### About Us
- [x]  about us bio modals have two close Xs at the top, remove the FA icon one

#### Footer
- [x] social icon hover color should match text color of the other footer elements
 - [x] "partner with us" button in sponsors section isn't consistent with other buttons

#### Buttons and Button components
- [x] newly added theme button classes don't have the exact :hover/:active colors we want

#### Alert header
- [x] Alert header close icon and text can push together with no spacing
- [x] alert header still uses white font on orange, fix the font color
 - [x] alert header margin was all margin-bottom instead of a split top/bottom to center in the background
- [x] change alert-header-close to min-width: 30px instead of fixed width 30px;


PeterBreen:
known issues:
 - [ ] "partner with us" button in sponsors section isn't consistent with other buttons
 - [ ] all footer buttons have not been changed
 - [ ] about us bio modals have two close Xs at the top, remove the FA icon one
 - [ ] user profile page links are blue (not orange) and need padding at the bottom of the section
 - [ ] user profile page has inline margin values that should be replaced with `col-xs-12 col-md-?` classes
 - [ ] create project "back" and "next" buttons aren't aligned in the same row properly
- [ ] donate page buttons are all wrong (outline black text, orange on hover - maybe add fill on select?)
- [ ] thank you page needs padding between image and footer

common/components/controllers/AboutUserController.jsx
```diff
@@ -49,35 +49,19 @@ class AboutUserController extends React.PureComponent<{||}, State> {
...
-                      </div>
-                    </div>
-                  </div>
-                </div>
-              </div>
-              <div className="col">
-              </div>
-              <div className="col col-sm-3">
-              </div>
+        <div className="container">
```
PeterBreen: As you can see from the simplification across this entire file I'm letting Bootstrap do more of our style lifting on this page. It's a change from how it used to display but not a redesign per se; I look at it more like falling in line with the rest of the site. 

Still, because it does involve a UI change, it should get some special attention on review to make sure we're all okay with it until the profile page gets to the top of design's queue.

civictechprojects/static/css/_bootstrapoverride.scss
```diff
@@ -28,9 +28,12 @@ $font-family-sans-serif: 'Montserrat',-apple-system, BlinkMacSystemFont, "Segoe
 
 //add DL colors to themes for button variants, etc
 $theme-colors: (
-  "dl-orange": #f79e02,
-  "dl-orange-darker": #de8e01
+  "primary": #f79e02, // use this for "action"
```
PeterBreen: Of note here are the primary, success, and danger keys - these are ones Bootstrap [needs to have](https://getbootstrap.com/docs/4.0/getting-started/theming/#required-keys) so we'll want to keep an eye out on buttons, forms, and input elements to make sure they're doing what we expect now that they've been rethemed.

common/components/common/projects/ProjectVolunteerModal.jsx
```diff
@@ -96,7 +96,12 @@ class ProjectVolunteerModal extends React.PureComponent<Props, State> {
   }
 
   onRoleChange(role: $ReadOnlyArray<TagDefinition>): void {
-    this.state.roleTag = role[0];
+    //this fixes an Uncaught TypeError: Cannot read property '0' of null if the user clicks the X to clear the field in the role selector
```
PeterBreen: I don't think this was added by my branch but I don't like console errors and it was a quick fix. when role was set to undefined or null, this component attempted to read null[0] which didn't work. Now it checks for that before setting state.

civictechprojects/static/css/partials/_AboutProject.scss
```diff
@@ -215,7 +215,10 @@
     width: 100%;
   }
 }
-
+.AboutProjects-jumplink {
+  /* anchor tag is moved "higher up" than its content to account for Bootstrap fixed nav bar */
```
PeterBreen: Because our nav bar is fixed, when we have a jumplink to a section on this page it will dutifully scroll so the anchor is at the top of the window - which is then hidden behind the nav. This is a no-JS trick to just let us 'pad' that out a bit, so our jumplinks will work as expected and the content will not be hidden behind the navigation.

common/components/common/projects/ProjectVolunteerModal.jsx
marlonkeating: Cool, you could make it a one-liner with 
```suggestion
   this.state.roleTag = !_.isEmpty(role) ? role[0] : null;
```

PeterBreen: Just made this change. Good catch, thanks!

civictechprojects/static/css/partials/_Global.scss
```diff
@@ -3,6 +3,7 @@ body {
 }
 
 // uses democracyLab orange for a "themed button", use class="btn btn-theme"
+// TODO: Use this for footer button concept, but remove this before merging reactstrap4 branch
```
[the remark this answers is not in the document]
PeterBreen: Yes, thank you. I didn't remove this old CSS I was using as reference. Rather than btn-theme we should use btn-primary, now.

PeterBreen:
Updated with changes from the review.

PR #298, opened 2019-11-14
Peter/2019-11 version updates

PeterBreen:
This PR is a whole batch of version upgrades for performance/security/etc. 

This branch should be merged with #295 for a full regression test prior to either one being deployed to prod

## Updates

To clear the checklist there were three types of upgrades: run with pip, run with yarn, and indirect dependencies, handled with "resolutions" in package.json. A list of all upgrades is below.

### pip
Updates with pip, which should be represented in the requirements.txt:

pip install psycopg2-binary
pip install django==1.11.22
pip install --upgrade djangorestframework

### yarn
Updates with yarn upgrade packagename --latest, which should be represented in package.json: 
lodash
node-sass

### resolutions
Some of our dependencies are not directly represented in our package.json, so I had to do a little digging.

Running `yarn why <packagename>` gives us this:
 - fstream: "node-sass#node-gyp" depends on it
 - sshpk: "node-sass#request#http-signature" depends on it
 - mixin-deep: "chokidar#braces#snapdragon#base" depends on it (chokidar is required by Webpack)
 - set-value: "chokidar#braces#snapdragon#base#cache-base" depends on it
 - js-yaml: "react-select#emotion#babel-plugin-emotion#babel-plugin-macros#cosmiconfig" depends on it

However, upgrading dependencies of dependencies is somewhat complicated, so I found yarn resolutions: https://yarnpkg.com/lang/en/docs/selective-version-resolutions/

And then added the "resolutions" field to package.json with the appropriate versions. Running `yarn install` afterwards updated yarn.lock, so it seems to have worked.

## notes
 After I submitted this PR, dependabot put in a whole series of dependency bumps, so I reviewed my changes and made sure to match dependabot in every area where I was behind. This PR now matches/exceeds all of PRs 299 - 305.


PeterBreen:
My django console has one warning to resolve:
> WARNINGS:
?: (rest_framework.W001) You have specified a default PAGE_SIZE pagination rest_framework setting,without specifying also a DEFAULT_PAGINATION_CLASS.
	HINT: The default for DEFAULT_PAGINATION_CLASS is None. In previous versions this was PageNumberPagination. If you wish to define PAGE_SIZE globally whilst defining pagination_class on a per-view basis you may silence this check.

docs: https://www.django-rest-framework.org/api-guide/pagination/#pagenumberpagination

PR #310, opened 2019-12-05
Long project urls truncated with ellipsis

Arpita-22:
Code changes to truncate long project urls with ellipsis.


PeterBreen:
> The change looks good. However, we prefer to put styling into separate scss files. I would put it in civictechprojects/static/css/partials/_AboutProject.scss

I agree with Marlon - it looks good. I have one question, though:

Did you try applying these style rules to the parent `<p>` element instead of the `<a>`? If not, it might be worth a try, because I believe it will work without having to declare `display: block` and `max-width: 250px`. The max-width: 250px is particularly troublesome on mobile, as it truncates the link before it needs to:

Screen Shot 2019-12-05 at 3 14 24 PM

Arpita-22:
> > The change looks good. However, we prefer to put styling into separate scss files. I would put it in civictechprojects/static/css/partials/_AboutProject.scss
> 
> I agree with Marlon - it looks good. I have one question, though:
> 
> Did you try applying these style rules to the parent `<p>` element instead of the `<a>`? If not, it might be worth a try, because I believe it will work without having to declare `display: block` and `max-width: 250px`. The max-width: 250px is particularly troublesome on mobile, as it truncates the link before it needs to:
> 
> Screen Shot 2019-12-05 at 3 14 24 PM

Thank you Marlon, Peter !!
Applying the styles to <p> did work out :) Moved the style to _AboutProject.scss.
The style is applied only to project urls though.

PR #315, opened 2020-01-01
List github commits for projects

marlonkeating:
This change adds a process for ingesting github commits for a project and displaying them on the project profile.

## Features ##
- Summary info on the latest commits from repositories covered by the project's code repository github link will be stored in the database
- Latest 5 commits shown on project profile page.  More can be loaded by clicking 'Show more activity'
- If the latest git commit is more recent than a project's updated date, date is updated to match commit date

## Supported Github Links ##
- Repository (example: https://github.com/DemocracyLab/CivicTechExchange)
- Organization (example: https://github.com/DemocracyLab/CivicTechExchange)
- User (example: https://github.com/marlonkeating)

## Configuration ##
- GITHUB_API_TOKEN should be a github api token.  Without it, git requests will be rate-limited to a prohibitively low number.
- MAX_COMMITS_PER_PROJECT is the maximum number of commits per project we will store in the database


democracylab/settings.py
```diff
@@ -90,6 +90,8 @@
     },
 }
 
+GITHUB_API_TOKEN = os.environ.get('GITHUB_API_TOKEN', None)
```
PeterBreen: I'd suggest adding a warning to ENVIRONMENT_VARIABLE_WARNINGS for this key about the rate limiter so (for example) a developer will know the github pulls may not work properly if it's not set in a dev environment.

common/management/commands/project_external_updates.py
```diff
@@ -0,0 +1,92 @@
...
+        print('Ingesting: ' + repo_url)
+        repo_info = fetch_github_info(repo_url)
+        if repo_info is not None and len(repo_info) > 0:
+            repo_display_name = repo_name[0] + '/' + repo_name[1]
+            raw_commits = raw_commits + list(map(lambda commit: [repo_display_name, commit], repo_info))
+
+    if len(raw_commits) > 0:
+        # Take the most recent top X commits
+        raw_commits.sort(key=lambda commit: commit[1]['commit']['author']['date'], reverse=True)
+        raw_commits = raw_commits[:settings.MAX_COMMITS_PER_PROJECT]
```
PeterBreen: Out of curiosity if we didn't set MAX_COMMITS in the settings what'd happen here? Might be worth adding a hardcoded value to use in case max commits doesn't exist?

marlonkeating: We do, the default is 30

PR #317, opened 2020-01-06
Peter/sponsor update

PeterBreen:
Updates our sponsors footer and Partner With Us page to split our sponsors into two categories, Sustaining and Advancing. Has commented out support for the other two categories in our sponsorship documentation.

In order for this to work, the SPONSORS_METADATA environment variable needs to be updated; each object in the array needs a `category` key added with a string value that matches their category name (e.g. `"Sustaining"` or `"Advancing"` and yes capitalization matters for now) which will slot them into the correct grouping. See this PR's environment variable file for a functioning copy of our current sponsors.

Styling is improved, but will probably need a real once-over when we get a more robust category system in place.

## known issues
- [x] sponsor section headers render even when no sponsor data exists
- [x] individual sponsor cards need vertical padding when stacked (on mobile, or with 3+ in a given category)
- [x] text padding on left edge for Partner With Us page (desktop only)
- [x] resize logo images to match on footer and PWU page
- [x] container padding for PWU Partners section needs refactor, mobile fix


common/components/chrome/MainFooter.jsx
```diff
@@ -36,33 +36,44 @@ class MainFooter extends React.Component<{||}> {
...
+        <div className="MainFooter-item col-12 text-center">
           <h2>Sponsors Make It Possible</h2>
           <p>Support the acceleration of social change</p>
           <a className="btn btn-primary" href={url.section(Section.PartnerWithUs)}>
             PARTNER WITH US
           </a>
         </div>
-        {this._renderSponsors()}
+        <div className="MainFooter-sponsor-container col-12">
+            {/* {this._renderSponsors("Visionary")} */}
```
marlonkeating: Since the code is here let's leave the unused tiers uncommented so we can add them via config later.

Issue #316, opened 2020-01-04
Inactive Slack link: README.md and CONTRIBUTING.md


aminamos:
Slack invite hyperlinks in README.md and CONTRIBUTING.md lead to an inactive destination, but the Slack invite from https://www.democracylab.org/index/?section=AboutProject&id=1 is valid.

PeterBreen:
Thanks for letting us know! As you can see I've put in a PR to update our documentation. Please let me know if the invite links in the docs work properly once we merge that in. :)

PR #319, opened 2020-01-09
update slack links in docs

PeterBreen:
Updates our README and CONTRIBUTING documents to have the same Slack invite link as our website, addressing #316 (thanks for letting us know about it!)

It's difficult for me to test since I'm a member of the slack already but the link seems to function so ... probably good?


aminamos:
It works, thank you!

PR #325, opened 2020-01-29
Bug Fix:  Prerender urls matching sitemap by substring rather than exactly

marlonkeating:
The way we were detecting whether prospective urls matched what was in the sitemap, and therefore should be prerendered, made it so that urls only had to match by substring, which led to many false positives.  This corrects the issue, and adds unit tests to test the changes with valid and invalid urls.

### DIFF
--- a/common/tests/test_helpers.py
+++ b/common/tests/test_helpers.py
@@ -1,0 +1,21 @@
+from django.test import TestCase
+from common.helpers.caching import is_sitemap_url
+from civictechprojects.sitemaps import all_sitemap_paths
+
+
+class CommonHelperTests(TestCase):
+
+    def test_prerender_urls(self):
+        urls = all_sitemap_paths
+        for url in urls:
+            self.assertTrue(is_sitemap_url(url), 'Should be able to prerender ' + url)
+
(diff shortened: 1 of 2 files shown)


common/tests/test_helpers.py
```diff
@@ -0,0 +1,21 @@
...
+
+    def test_prerender_urls(self):
+        urls = all_sitemap_paths
+        for url in urls:
+            self.assertTrue(is_sitemap_url(url), 'Should be able to prerender ' + url)
+
+    def test_do_not_prerender_urls(self):
+        urls = [
+            '/projects/signup/',
+            '/index/?section=FindProjects&sortField=project_name'
```
PeterBreen: If we're disallowing the sortField for by name ascending/descending, should we also disallow sort by location, e.g. `/index/?section=FindProjects&location=Seattle, WA` ?

marlonkeating: Yes, that's an equivalent example.  I opted to use examples for a backend url and a front-end url matching part of a sitemap url, but let me know if you can think of any other url categories we could include.

PR #329, opened 2020-02-17
Add Volunteering terms to Sign Up page

marlonkeating:
When new users come to sign up, they will now be asked to read and accept terms of service first.


common/components/controllers/SignUpController.jsx
```diff
@@ -72,6 +80,10 @@ class SignUpController extends React.Component<Props, State> {
           checkFunc: (state: State) => state.password1 === state.password2,
           errorMessage: "Passwords don't match"
         },
+        {
+          checkFunc: (state: State) => state.didCheckTerms,
+          errorMessage: "Check terms of service"
```
PeterBreen: "Agree to terms of service" seems a little more natural than "Check" (which could be check the box, or check them out)

PR #332, opened 2020-02-20
Validate URL inputs on Project Info form and Project Resource form

hoangviet1993:
- Whenever user enter an URL on the Project Info form or the Project Resource form, the URL will be validated immediately. User cannot proceed with creating new project with invalid URL.
- URL field on Project Info form can be empty.
- URL field on Link Entry modal (from Project Resource form) cannot be empty.

### DIFF
--- a/common/components/componentsBySection/CreateProject/ProjectInfoForm.jsx
+++ b/common/components/componentsBySection/CreateProject/ProjectInfoForm.jsx
@@ -11,7 +11,7 @@ import type {TagDefinition, ProjectDetailsAPIData} from "../../../components/uti
 import {Locations} from "../../constants/ProjectConstants.js";
 import formHelper, {FormPropsBase, FormStateBase} from "../../utils/forms.js";
 import _ from "lodash";
-import url from "../../utils/url";
+import url from "../../utils/url.js";
 
 
 type FormFields = {|
--- a/common/components/componentsBySection/CreateProject/ProjectInfoForm.jsx
+++ b/common/components/componentsBySection/CreateProject/ProjectInfoForm.jsx
@@ -5,11 +5,13 @@ import TagCategory from "../../common/tags/TagCategory.jsx";
 import TagSelector from "../../common/tags/TagSelector.jsx";
 import DjangoCSRFToken from "django-react-csrftoken";
 import {OnReadySubmitFunc} from "./ProjectFormCommon.jsx";
+import FormValidation from "../../../components/forms/FormValidation.jsx";
 import type {Validator} from "../../../components/forms/FormValidation.jsx";
 import type {TagDefinition, ProjectDetailsAPIData} from "../../../components/utils/ProjectAPIUtils.js";
 import {Locations} from "../../constants/ProjectConstants.js";
 import formHelper, {FormPropsBase, FormStateBase} from "../../utils/forms.js";
 import _ from "lodash";
+import url from "../../utils/url";
 
 
 type FormFields = {|
@@ -24,6 +26,7 @@ type Props = {|
   project: ?ProjectDetailsAPIData,
   readyForSubmit: OnReadySubmitFunc
 |} & FormPropsBase<FormFields>;
+
 type State = {|
   formIsValid: boolean,
   validations: $ReadOnlyArray<Validator>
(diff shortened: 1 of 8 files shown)


hoangviet1993:
@marlonkeating Can you let me know if changes in this PR are in line with what you are looking for with regards to URL validation?

common/components/test/utils.test.js
```diff
@@ -79,7 +79,11 @@ describe('utils', () => {
 
   	const args = urlHelper.arguments('/api/test?query=test&page=1');
   	expect(args['query']).toEqual('test');
-  	expect(args['page']).toEqual('1');
+    expect(args['page']).toEqual('1');
+    
+    expect(urlHelper.isEmptyStringOrValidUrl('')).toEqual(true);
+    expect(urlHelper.isEmptyStringOrValidUrl('google.com')).toEqual(true);
+    expect(urlHelper.isEmptyStringOrValidUrl('localhost:3000')).toEqual(false);
```
marlonkeating: Here are some others to try, based on what we've seen people input into the links field.
### True ###
http://www.unsecure.com
https://www.secure.com
https://multiple.parts.com/subdir
https://hyphenated-domain.next.com/
https://www.gnarly.com/url/viewer.html?&something=d38d408127d64407a7627f8e990908fe&view=388155.63,109.56,-69891.37,388028.5,247.89,-70066.36,0.95&lyr=1,1,1,1,1,1,1,1,0,0,1,1,1,1,1,1,1,1,1,1,1,1,1,1&wkid=2926&v=2
### False ###
N/A
TBD
Not built yet

common/components/componentsBySection/CreateProject/ProjectInfoForm.jsx
```diff
@@ -5,11 +5,13 @@ import TagCategory from "../../common/tags/TagCategory.jsx";
 import TagSelector from "../../common/tags/TagSelector.jsx";
 import DjangoCSRFToken from "django-react-csrftoken";
 import {OnReadySubmitFunc} from "./ProjectFormCommon.jsx";
+import FormValidation from "../../../components/forms/FormValidation.jsx";
 import type {Validator} from "../../../components/forms/FormValidation.jsx";
 import type {TagDefinition, ProjectDetailsAPIData} from "../../../components/utils/ProjectAPIUtils.js";
 import {Locations} from "../../constants/ProjectConstants.js";
 import formHelper, {FormPropsBase, FormStateBase} from "../../utils/forms.js";
 import _ from "lodash";
+import url from "../../utils/url";
```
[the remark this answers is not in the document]
hoangviet1993: Done!

common/components/test/utils.test.js
hoangviet1993: I have added these test cases to the validator tests.

common/components/forms/LinkEntryModal.jsx
```diff
@@ -28,21 +35,41 @@ class LinkEntryModal extends React.PureComponent<Props,State> {
 
   constructor(props: Props): void {
     super(props);
+    const formFields: FormFields = {
+      project_link: "",
+    };
+    const validations: $ReadOnlyArray<Validator<FormFields>> = [
+      {
+        checkFunc: (formFields: FormFields) => url.isValidUrl(
+          formFields["project_link"]),
```
marlonkeating:
```suggestion
        checkFunc: (formFields: FormFields) => formFields["project_link"] && url.isValidUrl(
          formFields["project_link"]),
```

We also need to check for blank url's here

hoangviet1993: isURL does not accept blank URL. Let me re-add that test case to the test suite.

marlonkeating: Actually I'm not sure whether formFields["project_link"] starts out as a blank string or null, but I do know is that in its current state the Link Entry Modal will let me submit a blank url, which is not something we want.  However you want to handle the check is fine; you might leverage _.isEmpty to check for both blank/null in this case.

hoangviet1993: There is a little race-condition here that I identified.
- Event 1: setState re-assigning formFields with `""`
- Event 2: formIsValid gets updated when modal is reset (i.e closing and opening the modal)
Previously, when the form is closed, setState did not get around to empty out formFields before the modal is open again, thus formIsValid still remains true with the previous formField value.
With my latest commit, whenever the modal gets reset, the formIsValid is updated **after** setState has finished re-assigning formFields.
Thanks for catching this :)

common/components/forms/LinkEntryModal.jsx
```diff
@@ -68,13 +110,20 @@ class LinkEntryModal extends React.PureComponent<Props,State> {
...
     //Sanitize link
-    this.state.linkInfo.linkUrl = url.appendHttpIfMissingProtocol(this.state.linkInfo.linkUrl);
-
-    this.props.onSaveLink(this.state.linkInfo);
-    this.close();
+    this.state.formFields.project_link = url.appendHttpIfMissingProtocol(
+      this.state.formFields.project_link);
+    this.setState({
+      linkInfo : {
+        linkUrl: this.state.formFields.project_link
```
marlonkeating: Need to save linkName as well

hoangviet1993: Done!

PR #337, opened 2020-03-02
Project Position form: Validate URL input

hoangviet1993:
- URL input field must be empty or valid URL.

### DIFF
--- a/common/components/forms/PositionEntryModal.jsx
+++ b/common/components/forms/PositionEntryModal.jsx
@@ -7,8 +7,16 @@ import {TagDefinition} from "../utils/ProjectAPIUtils.js";
 import TagSelector from "../common/tags/TagSelector.jsx";
 import TagCategory from "../common/tags/TagCategory.jsx";
 import {PositionInfo} from "./PositionInfo.jsx";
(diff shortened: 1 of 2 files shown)


common/components/forms/PositionEntryModal.jsx
```diff
@@ -7,17 +7,30 @@ import {TagDefinition} from "../utils/ProjectAPIUtils.js";
 import TagSelector from "../common/tags/TagSelector.jsx";
 import TagCategory from "../common/tags/TagCategory.jsx";
 import {PositionInfo} from "./PositionInfo.jsx";
-import url from "../utils/url.js";
-import _ from 'lodash'
+import FormValidation, {Validator} from "../forms/FormValidation.jsx";
+import formHelper from "../utils/forms.js"
+import urlHelper from "../utils/url.js";
+import isEmpty from 'lodash/isEmpty';
```
marlonkeating: Our convention has been to import _ and use _.isEmpty.  While your way here is perfectly valid, I feel we should keep to the existing pattern to help readability.

hoangviet1993: Done!

PR #343, opened 2020-03-05
OAuth login links re-skin

marlonkeating:
- OAuth service login icons have been replaced by colored icons with accompanying text, in a vertical layout
- svg usage has been standardized to use the webpack import method (and old code removed)

### TODO ###

- [ ] Resize and add padding to google.svg


civictechprojects/static/css/partials/_LogInController.scss
```diff
@@ -88,3 +88,20 @@
   height: 55px;
   cursor: pointer;
 }
+
+.LogInController-google svg{
+  width: 40px;
+  height: 50px;
+}
```
PeterBreen:
```suggestion
.LogInController-google svg {
  padding: 5px 0 4px 0;
}
```
Changing the svg size itself changes the entire size of the containing box (since the 55px SVG height effectively sets the height, as the tallest element) so instead let's pad the SVG to 'push' the image into the shape we want while retaining the 55px height. This prevents the Google box from being 5px shorter than Github, LinkedIn, etc.

civictechprojects/static/css/partials/_LogInController.scss
```diff
@@ -88,3 +88,20 @@
...
+
+}
+
+.LogInController-github svg{
+
+}
+
+.LogInController-linkedin svg{
+
+}
```
PeterBreen: Code cleanliness, suggest removing all of this. If we need to specify a rule for fb/git/linkedin we can do so.

civictechprojects/static/css/partials/_LogInController.scss
```diff
@@ -65,15 +65,43 @@
 }
 
 .LogInController-socialLink {
-  margin: 0 10px;
   background-color: transparent;
   text-decoration: none;
+  border: thin solid;
```
PeterBreen:
```suggestion
  border: 1px solid $color-background-dark;
```
border thin isn't guaranteed to be cross-browser consistent, so this is a minor "just in case" fix.

civictechprojects/static/css/partials/_LogInController.scss
```diff
@@ -65,15 +65,43 @@
...
+  margin-bottom: 10px;
+  width: 325px;
+  margin-left: auto;
+  margin-right: auto;
+
+  a {
+    color: $color-text-dark;
+  }
```
PeterBreen:
```suggestion
  a {
    color: $color-text-dark;
    padding-left: 15px;
  }
```

Once we remove the text-center class, this will help line up the icons nicely.

common/components/common/integrations/SocialMediaSignupSection.jsx
```diff
@@ -33,7 +55,7 @@ class SocialMediaSignupSection extends React.Component<{||}, State> {
   _renderSocialLoginsSection(): React$Node {
     return (
       <div className="text-center">
```
PeterBreen: I can't do a multiline suggest since this bit includes a deleted line, but I suggest this:
```
return (
  <React.Fragment>
    <h5 className="text-center">OR</h5>
    {this._renderSocialLogins()}
  </React.Fragment>
)
```
It does two things: Removes a div we don't need out of the DOM, and only applies text-center where it matters (aka the H5) which lets us line up our content in the SocialLogins more cleanly. Combined with the changes in the CSS to the `a` tag, it should line up both social icons and the "Continue with..." text cleanly.

PeterBreen:
There are some issues on mobile that are not breaking, just a little ugly. It's up to you if you want to take this on well after what should've been code freeze. I think we'll have to do them eventually.

Change 
`.LogInController-socialLink { width: 325px }` to max-width: 325px;

and then add  `display: flex; align-items: center;` to `.LogInController-socialLink a`
(because the a element is the parent for the two spans (svg logo and text) that's where we need the flex container to go)

It cleans it up just that little bit more and especially on narrow phones prevents either awkward scrolling or ugly line breaking.

PR #350, opened 2020-03-12
Single Project Card layout

PeterBreen:
Right now the DL website has different project card layouts for the homepage than find projects page (and the FP cards change layout on viewport size) - this branch is an experiment at standardizing on one layout for every page and viewport.

### DIFF
--- a/civictechprojects/static/css/partials/_ProjectCard.scss
+++ b/civictechprojects/static/css/partials/_ProjectCard.scss
@@ -8,7 +8,7 @@
 }
 .ProjectCard-root h2 {
   font-size: 1.333rem;
-  margin-top: 0.5rem;
+  margin: 0.5rem 0.5rem 0.5rem 0; /* left margin is built into card design */
 }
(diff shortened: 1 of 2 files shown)


civictechprojects/static/css/partials/_ProjectCard.scss
```diff
@@ -94,50 +95,3 @@
   padding: 0.25rem 0.5rem;
   margin: 0.25rem 0.5rem 0.25rem 0;
```
PeterBreen: this should be `.ProjectCard-title h2` in order to work properly (and the values should be changed to have 0.5 rem margin or padding, not both, top/right/bottom (left is provided by the grid layout having a spacer column, but if that's removed just make it margin: 0.5rem; or whatever)

to avoid this text hard up on the right edge: 
image

PeterBreen: I don't know why I attached that comment to the wrong line, but I fixed this in 9de94c7

PR #357, opened 2020-03-24
configured ImageCropUploadButton crosshairs to use square aspect rati…

lzapson:
…o or props aspect ratio


marlonkeating:
The Create Project and Edit Project thumbnail pickers are now locked to square aspect ratios, and we want to keep them flexible.

PR #360, opened 2020-04-02
Drap and drop image selector ui improvements

lzapson:


common/components/common/upload/ImageCropUploadButton.jsx
```diff
@@ -92,6 +97,15 @@ class ImageCropUploadButton extends React.PureComponent<Props, State> {
 
   _handleFileSelection(file): void {
     this.setState( { src: file, isCropping: true } );
+
+    // set cropping aspect ratio for selected image     
+    const image = new Image();
+    image.src = file;
+    image.onload = ({ target }) => {
+      if (!this.props.aspect && target.width && target.height) {
+        this.setState({crop: {...this.state.crop, aspect: target.width / target.height }})
```
marlonkeating, line 106: We don't need to lock the crop window to the image dimensions.

marlonkeating:
When you load up an image now, you can't see the cropping frame until you click on the image.  There are currently no visual cues to clue the reader into this.

marlonkeating:
Working better, but one other thing I'm noticing.  You can't crop all the way to the edge of the images now.  There's also wonky behavior where clicking and dragging the frame to the far right of the image, snaps it back to the left side of the image.

image

lzapson:
The min-width and min-height seem to cause the wonky behavior, I removed both, I thinks it's OK for cropping small images since I gave some padding to outside div. Please let me know if OK. 
I removed the padding on the ReactCrop, this was causing the problem with cropping all the way to the edge. 
I also added border radius to the drag target area I previous added, it was showing the square edge under the image.

PR #362, opened 2020-04-08
Navigation Update

PeterBreen:
Updates our site navigation, both in terms of what's displayed and how (categories, showing our IA) and in terms of code (removing one of the last Material UI holdouts, so we can remove it from our dependencies)

## To do items

 - [x] get final links for Groups and Events, add to nav
 - [x] add user avatar when in a logged-in state 
 - [x] better signaling between 'this is a link' and 'this is a dropdown menu' 
 - [x] decide if we want hard href links or to reuse the Dispatcher 
- [x]  user nav dropdown (desktop) instead of list (e.g. my projects, my profile, log out)
- [x] user nav not in a separate dropdown (mobile) but in a list
- [x] add a 'default avatar' to replace MaterialUI's `<Person />`
- [x] fix repeated click of Log In link leading to `&prev=LogIn&prev=LogIn&prev=LogIn` querystring
- [x] stretch goal: move Log In and Log Out links to bottom of the mobile menu as they currently are 
- [x] generate links with helper method so we can highlight active section
- [x] add (configurable) link to DemocracyLab blog
 - [x] (mobile) indent menu text from left edge
 - [x] (mobile) test whether or not we should remove the box border around an expanded dropdown
 - [x] (both) move the dropdown arrow indicator down so it's not top-aligned
 - [x] (desktop) change Donate button to link 
 - [x] (desktop) change Log In link to button; black text with orange border (this helps it stand out from the alert header on mobile and doesn't contradict accessibility rules in our style guide, though we don't have specifically defined outline buttons)
 - [x] (mobile) change mobile-specific Log In button to black text/orange border, as above
 - [x] (both) change default font from grey to our $color-text-dark;
 - [x] (desktop) spacing on left edge of logo where there's room 
 - [x] (mobile) 320px display modifications so logo, log in, and menu still all fit in one line


common/components/controllers/LogInController.jsx
```diff
@@ -23,7 +23,7 @@ class LogInController extends React.Component<Props, State> {
     this.state = {
       username: '',
       password: '',
-      prevPage: window.location.href.split('&prev=')[1] || this.props.prevPage || document.referrer.split('section=')[1] || ''
+      prevPage: window.location.href.split('&prev=')[1] || this.props.prevPage || ''
```
PeterBreen: One nice thing about document.referrer we might want to keep in mind for the future is that it required passing no props or additional information in the query string, it 'just worked' (or didn't) automatically. 

Checking the PROTOCOL_DOMAIN env key to ensure only internal referrers are parsed here may be valuable if we choose to re-enable this case down the line - though it's not required at this time, since we changed how the Log In links are constructed in the nav to begin with.

common/components/chrome/MainHeader.jsx
```diff
@@ -85,307 +52,129 @@ class MainHeader extends React.Component<{||}, State > {
...
+        {CurrentUser.isLoggedIn() ? null : <Button className="MainHeader-showmobile MainHeader-login-button" variant="outline-primary" href={url.section(Section.LogIn, {prev: this._generatePrev()})}>Log In</Button>}
+        <Navbar.Toggle aria-controls="nav-pagenav-container" />
+        <Navbar.Collapse id="nav-pagenav-container" className="flex-column">
+          <Nav className="MainHeader-usernav ml-auto">
+            {CurrentUser.isLoggedIn() ? this._renderUserSection() : this._renderLogInSection()}
+          </Nav>
+          <Nav className="MainHeader-pagenav mr-auto">
+            <Nav.Link href={url.section(Section.Home)}>Home</Nav.Link>
+            <NavDropdown title="Projects" id="nav-projects">
+              <NavDropdown.Item href={url.section(Section.FindProjects)}>Find Projects</NavDropdown.Item>
```
marlonkeating, line 84: One thing we could do to more cleanly encapsulate conditional rendering for highlighting the current section is use a helper method for generating the Item.

renderNavDropdown(section, text) {
  // Return NavDropdown.Item that adds class with highlight if section == current section
}

PeterBreen: Added this in 8db706c though it could still use some further work.

democracylab/settings.py
```diff
@@ -402,4 +402,6 @@ def read_connection_config(config):
 
 BOARD_OF_DIRECTORS = os.environ.get('BOARD_OF_DIRECTORS', '')
 
-FAVICON_PATH = os.environ.get('FAVICON_PATH', 'https://d1agxr2dqkgkuy.cloudfront.net/img/favicon.png')
\ No newline at end of file
+FAVICON_PATH = os.environ.get('FAVICON_PATH', 'https://d1agxr2dqkgkuy.cloudfront.net/img/favicon.png')
+
+BLOG_URL = os.environ.get('BLOG_URL', 'https://blog.democracylab.org')
```
PeterBreen, line 407: Let me know if this is how we want it to fall back if no BLOG_URL is declared manually. My thinking was we could always retarget the blog.democracylab.org cname record to go wherever we want so it'd be our backup option and the blog link would always be visible and working. On the other hand if we want this to be a "show blog link _only if_ BLOG_URL is declared, otherwise hide it entirely" I'd have to rewrite this.

common/components/chrome/MainHeader.jsx
```diff
@@ -85,307 +52,139 @@ class MainHeader extends React.Component<{||}, State > {
...
-              {link.name}
-            </div>
-          </a>
-          <Divider />
-        </React.Fragment>
-      );
-    });
+//TODO: Refactor these to reduce duplication
+//TODO: Allow multiple arguments for url.section to handle url.getPreviousPageArg() - options object?
+  _renderNavLink(sec, text, classes = "") {
```
marlonkeating, line 123: nit: "sec" isn't a universally understood shorthand for 'section', and it only saves 4 characters, so I'd just use 'section' here

common/components/chrome/MainHeader.jsx
```diff
@@ -85,307 +52,139 @@ class MainHeader extends React.Component<{||}, State > {
...
+    // example: this._renderNavLink("Section.FindProjects", "Find Projects")
+    const urlArgs = url.arguments(url.section(sec))
```
marlonkeating, line 125: Missing semicolons here and in much of the following cod.e.

democracylab/settings.py
marlonkeating: Actually let's not do this default blog url, since there might be cases where we want the blog to be taken off entirely.

civictechprojects/static/css/partials/_MainHeader.scss
```diff
@@ -1,76 +1,110 @@
 .MainHeader-root {
-  font-size: 12px;
-  height: 54px;
+  // height: 79px;
```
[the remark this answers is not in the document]
PeterBreen: Apparently not! I checked the git blame and it was part of an old attempt at manipulating the logo, I think? Either way, good catch.

civictechprojects/static/css/partials/_MainHeader.scss
```diff
@@ -1,8 +1,10 @@
 .MainHeader-root {
-  // height: 79px;
   z-index: 99;
   width: 100%;
   position: fixed; /* fix the entire component, including AlertHeader and Nav */
+  // height: 100vh;
```
marlonkeating, line 5: These commented for any reason?

PeterBreen:
Make sure to check the AlertHeader <a> color on hover, as it's currently orange on orange in dev, which is not intended.

common/components/chrome/MainHeader.jsx
```diff
@@ -53,437 +41,160 @@ class MainHeader extends React.Component<{||}, State > {
...
+              {this._renderNavDropdownItem(Section.FindEvents, "Find Events")}
+              {this._renderNavDropdownItem(Section.CreateEvent, "Create Event")}
+            </NavDropdown>
+            <NavDropdown title="About" id="nav-about">
+              {this._renderNavDropdownItem(Section.AboutUs, "About Us")}
+              {this._renderNavDropdownItem(Section.ContactUs, "Contact Us")}
+              {this._renderNavDropdownItem(Section.Press, "News")}
+            </NavDropdown>
+            {window.BLOG_URL ? <Nav.Link href={window.BLOG_URL}>Blog</Nav.Link> : null}
+            {window.EVENT_URL ? <Nav.Link href={_.unescape(window.EVENT_URL)}>Live Event</Nav.Link> : null}
```
marlonkeating, line 110:
```suggestion
            {window.EVENT_URL ? <Nav.Link href={_.unescape(window.EVENT_URL)}>Event</Nav.Link> : null}
```

Should just be 'Event' since we show this even when the event is not live.

common/components/chrome/MainHeader.jsx
```diff
@@ -53,437 +41,160 @@ class MainHeader extends React.Component<{||}, State > {
...
+          <Nav className="MainHeader-usernav ml-auto">
+            {CurrentUser.isLoggedIn() ? this._renderUserSection() : this._renderLogInSection()}
+          </Nav>
+          <Nav className="MainHeader-pagenav mr-auto">
+            {this._renderNavLink(Section.Home, "Home")}
+            <NavDropdown title="Projects" id="nav-projects">
+              {this._renderNavDropdownItem(Section.FindProjects, "Find Projects")}
+              {this._renderNavDropdownItem(Section.CreateProject, "Create Project")}
+            </NavDropdown>
+            <NavDropdown title="Events" id="nav-events">
```
marlonkeating, line 100: For now, we only want to show Create Event for admins.

common/components/chrome/MainHeader.jsx
```diff
@@ -53,437 +41,160 @@ class MainHeader extends React.Component<{||}, State > {
...
+    if (urlArgs.section === this.state.activeSection) {
+      classes += " MainHeader-active";
+    };
+    return <Nav.Link className={classes} href={url.section(section)}>{text}</Nav.Link>
   }
-
-  _toggleSlider(open: boolean): void {
-    this.setState({ slider: open })
+  _renderNavDropdownItem(section, text, classes = "") {
+    const urlArgs = url.arguments(url.section(section));
```
marlonkeating, line 139: This code snippet is duplicated with _renderNavLink.  Perhaps we can wrap it into a helper method?

democracylab_environment_variables.sh
```diff
@@ -121,3 +121,6 @@ export FAVICON_PATH=https://d1agxr2dqkgkuy.cloudfront.net/img/favicon.png
 export BLOG_URL='https://blog.democracylab.org'
 
 export PRIVACY_POLICY_URL='https://d1agxr2dqkgkuy.cloudfront.net/documents/2020.06%20DemocracyLab%20Privacy%20Policy.pdf'
+
+# if we have a hackathon or similar event we want displayed in the nav for the duration, put it here
+export EVENT_URL='https://democracylab.org/index/?section=LiveEvent&id=1'
```
marlonkeating, line 126:
```suggestion
export EVENT_URL='https://democracylab.org/index/?section=AboutEvent&id=1'
```

PR #367, opened 2020-04-22
add-transition-for-active-projects - for width >= 992, added carousel…

lzapson:
…, removed prev and next controls


common/components/componentsBySection/Landing/RecentProjectsSection.jsx
```diff
@@ -127,8 +90,30 @@ class RecentProjectsSection extends React.Component<{||}, State> {
...
+
+  _carouselSlidHandler(eventKey, direction): void {
+    if (eventKey === 1) {
+      document.querySelector('.carousel-control-next-icon').setAttribute("style", "display:none");
+      document.querySelector('.carousel-control-prev-icon').setAttribute("style", "display:inline-block");
+    } else if (eventKey === 0) {
+      document.querySelector('.carousel-control-prev-icon').setAttribute("style", "display:none");
+      document.querySelector('.carousel-control-next-icon').setAttribute("style", "display:inline-block");
+    }
+  }
```
PeterBreen, line 107: This handler hides the icons, but the click targets themselves are still visible and will still toggle the card carousel.
Screen Shot 2020-05-01 at 12 34 45 PM

I think retargeting from `.carousel-control-next-icon` to `.carousel-control-next` (and the same for prev) should fix that?

lzapson: Hi Peter, thanks for your feedback much appreciated. I adjusted which gave more width to project cards! I liked how you captured the resize, very cool. Please let me know what you think.

PeterBreen: It looks better! :) Thanks for the fast turnaround, sorry I'm so late in responding here.

Quick note: If I'm reading the react-bootstrap carousel code correctly, it should be switching  the prev/next controls from none to flex, rather than none to inline-block. 

I'm having an issue with the carousel arrow though, it always shows the next arrow and never the prev one, at least in my VM.  Not sure why, but it's like the eventKey isn't being passed right so the conditional never toggles in the `_carouselSlidHandler` function.

A couple thoughts that may/may not help: There are two carousels on the homepage so is the querySelector getting the right one? And second, do you think changing the SlidHandler function to operate like a toggle (as in `display === 'none' ? 'flex' : 'none'`) could be valuable? It'd mean no need to pass an event key, but it'd require getting the current display state directly and that may end up more complex than the current solution. Up to you.

lzapson: Thanks for your feedback, no worries on getting back to me, please get back to me when you can. 

I changed to display:flex, the inline-block was for the icon, good catch, ty. 

In regard to targeting the correct carousel on the page, I added a RecentProjects-carousel class to the carousel, so I can loop through children.

I liked the toggle idea but I was not able to get it to work because the getAttribute('style') returns null as the inline style was not previous set. I am still checking eventKey. I'm concerned that you are not getting eventKey. I updated the "react-bootstrap" package, did you run yarn install? If the carousel onSlid API function does not work for you I need to come up with another solution. Please let me know.

PeterBreen: yarn install fixed it, thank you. works great here 👍

lzapson: Fantastic! Thanks for the confirmation.

PR #374, opened 2020-05-27
Donate/ThankYou template update

PeterBreen:
Updates our Donate and Thank You pages to use the same layout/template as the rest of the site.

Done in tandem with #380, though not dependent on it.


marlonkeating:
I'm seeing some cutoff on the mobile views.  This was a slight problem on the existing pages, but seems to be exacerbated by this change.

image

PeterBreen:
> I'm seeing some cutoff on the mobile views. This was a slight problem on the existing pages, but seems to be exacerbated by this change.

I've just pushed a change which should fix this. The fixed-width "One-time" and "Monthly" buttons were not wrapping onto two lines at narrow viewports; so now they do that. As a consequence though, on mobile the page has a lot of buttons in mostly vertical order which necessitates a lot of scrolling and the lack of visual difference between the two button sections during that scrolling could cause issues. As an experimental change I added section headings to make it clearer there are two sections with the implication a user has to click one from each section. That can be removed if it's a problem.

There are also some layout fixes which I applied to both the  Donate and Thank You pages for consistency.

PeterBreen:
Brought this up to date with a master merge, which I did poorly (whoops) so I think it's cleaned up now but I'd appreciate another set of eyes on making sure I didn't accidentally delete a section or something.

marlonkeating (approved):

PR #379, opened 2020-06-14
Add Location Autocomplete

marlonkeating:
## Search By Location ##
-Add search fields where user can search by location and search radius
-User must select Country, and then can type in their address, city, or zip code into the Near field

## Project Location ##
-User can select country, and use Location Autocomplete field on the Create and Edit project pages


common/components/controllers/EditProfileController.jsx
```diff
@@ -5,6 +5,7 @@ import DjangoCSRFToken from 'django-react-csrftoken'
 import UserAPIUtils from "../utils/UserAPIUtils.js";
 import type {UserAPIData} from "../utils/UserAPIUtils.js";
 import {CountrySelector, defaultCountryCode} from "../common/selection/CountrySelector.jsx";
```
PeterBreen, line 7:
```
WARNING in ./common/components/controllers/EditProfileController.jsx 64:15-33
"export 'defaultCountryCode' was not found in '../common/selection/CountrySelector.jsx'
```

Webpack gives this warning on build for me. It might not be new?

common/components/constants/ProjectConstants.js
```diff
@@ -1,3 +1,4 @@
+// TODO: Delete?
```
PeterBreen, line 1: Confirm if we can delete this?

marlonkeating: The in-progress Groups code is still using it, so adding a note to delete it after that feature is finished.

PR #380, opened 2020-06-18
About / Homepage splash image updates

PeterBreen:
- [x] updates image in top splash on About Us
 - [x] updates SplashScreen component to accept opacity value as prop (defaults to 0.5 / 50%), updates some About Us images to use the component instead of being hand built
 - [x] a lot of text and positioning changes on homepage from design team, most notably repositioning text to be in the bottom third (ish) of splash images instead of centered, text/alignment fixes
- [ ] mobile text sizes for splash images (still under discussion)

This branch also needs CDN drop-in replacements for image on Donate and ThankYou pages, but does not alter those pages directly. See #374 for those changes.

### DIFF
--- a/civictechprojects/static/css/partials/_LandingController.scss
+++ b/civictechprojects/static/css/partials/_LandingController.scss
@@ -42,7 +42,9 @@
   margin: 5px; /* override this on desktop */
 }
 .LandingController-root .SplashScreen-content {
-  margin: auto;
+  padding-bottom: 2rem;
+  margin-left: auto;
+  margin-right: auto;
 }
 .LandingController-topsplash .SplashScreen-content h1 {
   font-size: 4rem;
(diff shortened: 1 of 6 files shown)


civictechprojects/static/css/partials/_LandingController.scss
```diff
@@ -42,10 +42,16 @@
...
-  margin: auto;
+  padding-bottom: 2rem;
+  margin-left: auto;
+  margin-right: auto;
+}
+.LandingController-root .SplashScreen-content h1 {
+  font-size: 3rem;
 }
 .LandingController-topsplash .SplashScreen-content h1 {
   font-size: 4rem;
```
PeterBreen, line 53: In case I'm unavailable to finish this item: For mobile text changes still under discussion, change them here, but first copy and paste these font-size rules as-is into the desktop media query starting on line 206.

PR #386, opened 2020-06-26
Add Error page with custom message for OAuth signup failure

marlonkeating:
## Summary ##
Some users are failing to sign up with OAuth because their profile does not have all the fields we require (email, first, and last name).  We want to alert the user to this problem.

## Front-end ##
- Basic Error page that reads specific errors from the back-end

## Back-end ##
- Custom exception class ReportableError, structured for publishing error details to logs and to the front-end
- Custom error 500 handler that intercepts ReportableErrors, logs them, and redirects to Front-end error page

## Notes ##
- The custom error page will only be triggered if DJANGO_DEBUG=False.  If Debug is true, Django will output a comprehensive debug page that is useful for debugging but a security risk in Production.

### DIFF
--- a/common/helpers/error_handlers.py
+++ b/common/helpers/error_handlers.py
@@ -1,0 +1,31 @@
+import sys
+from django.shortcuts import redirect
+from common.helpers.constants import FrontEndSection
+from common.helpers.dictionaries import merge_dicts
+from common.helpers.front_end import section_url
+
+
+class ReportableError(Exception):
+    """Exception raised that needs to be logged and sent to a front end error page
+
+    Attributes:
+        message -- explanation of the error to be reported in the logs
+        front_end_args -- arguments to be surfaced on the front end error page
+    """
+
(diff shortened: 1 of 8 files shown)


common/helpers/error_handlers.py
```diff
@@ -0,0 +1,31 @@
...
+        message -- explanation of the error to be reported in the logs
+        front_end_args -- arguments to be surfaced on the front end error page
+    """
+
+    def __init__(self, message, front_end_args):
+        self.message = message
+        self.front_end_args = front_end_args or {}
+
+
+def handle500(request):
```
saysanjeev, line 21: Just one question- Till now there was no explicit error handler page as such in Front-End. With this change, any 500 error is redirected to this new Error page. Are we fine with that?

marlonkeating: Yes, as even a bare-bones error page is better than the blank error 500 page people were getting up until now.

PR #392, opened 2020-07-10
apply appendHttpIfMissingProtocol() to new project url

hanuz06:
<a> tag's href requires http/https prepended. Applied appendHttpIfMissingProtocol() to new project url


.gitignore
```diff
@@ -110,5 +110,8 @@ package-lock.json
 # Docker environment config
 docker_environment_init.sh
 
+#Django environment variables
+democracylab_environment_variables.sh
```
marlonkeating, line 114: We do still want this file around as a sample.  To guard against accidental changing and committing, we can do something like add the following to the top of the .sh file to keep it from being run accidentally:

read -n 1 -r -s -p $'This is an example .sh file, and should not be run or changed unless new environment variables are added.  Press any key to exit\n'
exit 1

hanuz06: Added a protection code to democracylab_environment_variables.sh and removed democracylab_environment_variables.sh
 from .gitignore.

PR #390, opened 2020-07-07
Groups MVP

marlonkeating:
# Groups Feature #
Groups are intended to represent a civic tech group's presence on the platform.  For this first iteration, they mainly serve as a repository of projects owned by a single organization.

## Create Groups ##
- Link to Create Group on top bar
- User enters in group information in the Create Group workflow
- Group must be approved by admin before showing up on Find Groups page

## Browse Groups ##
- Link to Find Groups on top bar
- Can search for groups by issue area, location, and keyword

## Group Profile ##
- Displays group information, as well as listing all projects that are part of the group
- Lists issue areas based on the issue areas of the constituent projects

## Inviting Group to Project ##
- Group owner can invite the project via the 'Invite Project to Group' button on Project profile
- Project owners will receive email with invitation, and can approve via email
- On the project profile, project owners will see entries for pending group approvals, and can approve or reject the invitation there

PR #396, opened 2020-07-16
Groups/header

PeterBreen:
Requires #390 

Adds splash to Find Groups page.

open questions before merge:
- [ ] full width or constrained splash image? Compare to top splash on homepage vs middle splash on homepage, for examples.
- [ ] font sizes and button colors: These are completely different from homepage (all bold, header text at 80 vs 64, etc) - is there a standard to use?
- [ ] button actions: What does the Find Groups button do since we're already on the Find Groups page. Make the splash disappear (and remove the link to Create) or....???


PeterBreen:
We're not going this route for now so I'll close this to clean up our open PRs.

PR #402, opened 2020-07-28
Update project's project_date_modified field on any delete action

asherdale:
Update project's project_date_modified field on any delete action

### DIFF
--- a/civictechprojects/models.py
+++ b/civictechprojects/models.py
@@ -97,6 +97,7 @@ class Project(Archived):
 
     def delete(self):
         self.is_searchable=False
+        self.project_date_modified = timezone.now()
         super().delete()
 
     def all_owners(self):
--- a/civictechprojects/models.py
+++ b/civictechprojects/models.py
@@ -97,7 +97,7 @@ class Project(Archived):
 
     def delete(self):
         self.is_searchable=False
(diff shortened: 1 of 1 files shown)


civictechprojects/models.py
```diff
@@ -97,6 +97,7 @@ def __str__(self):
 
     def delete(self):
         self.is_searchable=False
+        self.project_date_modified = timezone.now()
```
marlonkeating, line 100: Want to make sure and save() as well.  You can use update_timestamp() method below, which does this.

asherdale: save() is called in the "super().delete()" call, so I decided to not call that method so we don't duplicate save() calls. Does that reasoning make sense, or should I still call that helper method?

asherdale: I figured this way makes more sense because self.is_searchable is also modified but relies on the save() in "super().delete()" rather than having it's own save() call

marlonkeating: Yes, that's a good point.  I think in this case we're okay saving twice, as deletes happen rarely enough that performance hits aren't a worry.  In my view, using the same method to update timestamps is helpful because it makes it easier to track later if we ever need to audit where we're updating project timestamps.

PR #432, opened 2020-09-03
Project events listing

marlonkeating:
## Project Profile ##
- Section listing all of project's past and future events
- Jumplink that scrolls down to events section

### DIFF
--- a/civictechprojects/static/css/partials/_AboutProject.scss
+++ b/civictechprojects/static/css/partials/_AboutProject.scss
@@ -230,10 +230,15 @@
 /* events section overrides from Event page style to About Project page style */
 .AboutProjects-events {
   h1 {
-    font-size: 1.5rem;
+    font-size: 28px;
+    margin-top: 2.5rem;
+  }
+  h1:first-of-type {
+    margin-top: 0;
   }
   .EventCard-dateheader h4 {
-    font-size: 1.25rem;
+    font-size: 24px;
+    margin-bottom: 1rem;
   }
 }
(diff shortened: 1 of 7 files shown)


civictechprojects/static/css/partials/_AboutProject.scss
```diff
@@ -227,10 +227,17 @@
     width: 100%;
   }
 }
-
-.AboutProjects-recent-activity {
-  padding: 20px; /*as with description-details consider maxwidth */
-
+/* events section overrides from Event page style to About Project page style */
+.AboutProjects-events {
```
PeterBreen, line 231: It's a bit awkward that `<EventCardsListings />` uses an H1 here (it's fine on the actual Events page) so we'll probably want to fix this for proper hierarchy purposes later, but I made it display as other section headings do.

civictechprojects/static/css/partials/_AboutProject.scss
```diff
@@ -227,10 +227,17 @@
...
+  h1 {
+    font-size: 1.5rem;
+  }
+  .EventCard-dateheader h4 {
+    font-size: 1.25rem;
+  }
+}
+.AboutProjects-recent-activity, .AboutProjects-events  {
```
PeterBreen, line 239: Just a quick change, this section was previously uniform 20px padding on all 4 sides, but the other sections on page are 25 top/bottom, 10 left/right, so now they should all do that including the new Events section.

PeterBreen (approved):

PR #436, opened 2020-09-11
Form Fixes and Polish

marlonkeating:
## Create/Edit Page ## 
- Form Next/PUBLISH buttons are now disabled while the save api call is ongoing
- Spinner is shown on button while the save api call is ongoing 
- Create Project workflow now used when editing project after publishing

## Caching ##
- Refresh project tag cache when admin approves project
- Only re-cache project tag counts when tags change for a searchable project upon saving

## Project Profile ##
- Don't show Contact/Volunteer buttons when in embedded viewa

### DIFF
--- a/civictechprojects/forms.py
+++ b/civictechprojects/forms.py
@@ -79,7 +79,8 @@ class ProjectCreationForm(ModelForm):
             print('notifying project creation')
             send_project_creation_notification(project)
 
-        Cache.refresh(CacheKeys.ProjectTagCounts)
+        if project.is_searchable:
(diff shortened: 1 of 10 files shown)


PeterBreen (approved):
I tried to do a lot of tag add/remove to get it to break, but it worked every time! That's good, right?

Also, we discussed this in standup, but upon review I think the UX for how it works while saving is just fine - the button has the progress spinner and the message `Saved!` appears next to the button when the save is complete. Those two combined work out well. I'm more than content to leave that as-is.

PR #437, opened 2020-09-14
Fix new user newsletter signup error

marlonkeating:
We weren't passing new users' first and last names when signing them up for our newsletter.


PeterBreen (approved):
First and last names successfully passed to the email list.

PR #438, opened 2020-09-17
Cache project json data

marlonkeating:
Our full project json contains a lot of data from multiple tables, contributing to the load time for the project profile page.  This change caches the project json, refreshing the data when the project or objects displayed on the project page(Groups, Events, Volunteers) change.

### DIFF
--- a/civictechprojects/caching/cache.py
+++ b/civictechprojects/caching/cache.py
@@ -1,0 +1,18 @@
+from common.caching.cache import Cache
+
+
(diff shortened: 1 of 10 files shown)


PeterBreen (approved):
Testing via dev tools I saw a noticeable increase in API performance (something like 40-45%, or a query that used to take 300ms now takes about 170ms) - and a small reduction in total "blocking time" as estimated by devtools, as well.

PR #446, opened 2020-10-03
Preload project headers

marlonkeating:
# Overview # 
Currently our head metadata is loaded in after initial load, as we load in project data after initial load.  This change leverages our new project cache to load in the project metadata headers during the initial page fetch.  

## Page-Specific Preload Handlers ## 
This framework can be easily extended to preload custom headers for other pages beyond the project profile.  Pages can be mapped to handlers, defaulting to standard DemocracyLab headers when specific pages do not have a custom handler.

## Included Bug Fixes ##
- Project cache entries now never expire


PeterBreen (dismissed):
Overall it's pretty good (opengraph fields are picked up, e.g. title and description) so it works and does seem to generate correct previews, with one exception: there's something going on with `og:image` that's not working right.

I tested a couple projects (e.g. Orcasound) on LinkedIn's [Post Inspector](https://www.linkedin.com/post-inspector/) and [Social Share Preview](https://socialsharepreview.com/) and for both of them og:image did not seem to work and the image was not shown.

If I go to the project page itself, and then inspect the HTML after the page has finished rendering shows a complete and (seemingly) correct meta property for the image -- but if I try to navigate to the image directly by copy/pasting the URL it immediately downloads rather than displaying in browser. Perhaps opengraph doesn't like that?

Anyway, it's pretty solid as is. If we're okay with og:image not working then I can approve, but I'd like to see if it can be solved.

common/helpers/front_end.py
```diff
@@ -16,3 +17,8 @@ def get_page_section(url):
     if match is not None:
         groups = match.groups()
         return groups[0]
+
+
+def get_clean_url(url):
```
PeterBreen, line 22: As a hotfix, this works (I tested it replacing `&amp;` and `&equals;` and it did both) though it fails on something like `&#61;` or `&#63` entity numbers.

I know in js unescape() has been [functionally deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/unescape) in favor of decodeURI, so I'm wondering if - for long-term use of this function  - we should consider `urllib.parse` or similar rather than just an unescape?

PeterBreen (approved):
After some discussion in standup we've decided to approve this as a hotfix (limited scope) and add two trello tickets for future improvements/enhancements.

common/helpers/front_end.py
marlonkeating: We can look into this some more, but I haven't seen any instances where entity numbers get into the url so that can be a future action item.

PR #451, opened 2020-10-31
Private Event Support

marlonkeating:
# Overview #
Private events are inaccessible to anyone who does not know the name of the events.

## Create Event Page ##
- Non-admins can now create events, but the link to the page is not public
- Admins can set whether an event is private, and the canonical name of the event
- Changing an event slugs to be a duplicate of another event will fail
- Now only admins can set the qiqochat id and legacy organization fields

## Event Profile Page ##
- Can now use event slug in place of the url &id argument
- For private events, only the &id=slug construction will work
- Bug Fix: Event RSVP link adds http: if not provided

## Create Project Page ##
- Now only admins can set the legacy organization tag field


civictechprojects/views.py
```diff
@@ -248,7 +248,14 @@ def event_delete(request, event_id):
 
 
 def get_event(request, event_id):
-    event = Event.objects.get(id=event_id)
+    if event_id.isnumeric():
```
PeterBreen, line 251: I think there's a potential source of confusion if `id=15` or `id=slugname` works for the owner - who would presumably want to send out links to their own event, and suddenly a link that works for them would fail silently for everyone else. Would it be possible to make it so that if an event has a slug ID, it is _always_ redirected to that slug ID for owners and visitors alike (admins could still use both if we need that), so there's no possible copy-and-paste failure state?

PeterBreen (changes requested):
One requested change, see comment. Every change listed here seems to operate as described, and if the requested change is going to be a struggle we can stick it on Trello for later.

civictechprojects/views.py
marlonkeating: Good point, we need a robust way to handle canonical urls.

PeterBreen: Okay, I saw the trello ticket for canonical. Works for me.

PeterBreen (approved):

PR #452, opened 2020-11-02
add recaptcha v3 instrumentation

PeterBreen:
So that we can set our recaptcha v3 thresholds correctly, this branch adds recaptcha V3 instrumentation, which will allow us to passively collect scoring information before setting thresholds and refactoring our contact form to use V3 over V2 as the anti-spam check.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # https://hub.docker.com/r/nikolaik/python-nodejs
-FROM nikolaik/python-nodejs:python3.8-nodejs12	
+FROM nikolaik/python-nodejs:python3.8-nodejs12
 
 # Replace shell with bash so we can source files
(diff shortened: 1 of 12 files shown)


PeterBreen:
It appears that I did not add enough instrumentation to get Recaptcha to produce any data despite running a bunch of traffic through the deployment both manually and with K6.io virtual users - I think I need more than just the frontend side to get anything out of recaptcha v3, despite the v3 badge appearing and suggesting otherwise. Most likely, setting up backend validation and configuring "actions" in the Recaptcha admin panel are needed as well - even for analytics, much less challenges. 

While I think that may be valuable in the future, the cost/benefit in dev time to get that fully implemented for one low-traffic contact form is not present at the moment. But I'll keep this around as reference. :)

PR #453, opened 2020-11-06
Qiqochat signup throttling

marlonkeating:
Sometimes two requests are sent in quick succession to sign a new user up to qiqochat, which can cause database corruption on qiqochat's side.  This change ensures that two calls aren't set in quick succession for the same user


PeterBreen (approved):
After discussing this with Marlon on Slack I'm comfortable with what we've got here.

PR #455, opened 2020-11-14
11/16/2020 Release Fixes

marlonkeating:
## Volunteer Email ##
We want to call out to project owners that they can email volunteers by replying to the email sent out for a volunteer applying to join.

## Event Card ##
- Event cards should link to the event slug instead of the event id if a slug is set

### DIFF
--- a/civictechprojects/models.py
+++ b/civictechprojects/models.py
@@ -431,6 +431,7 @@ class Event(Archived):
             'event_date_end': self.event_date_end.__str__(),
             'event_date_start': self.event_date_start.__str__(),
             'event_id': self.id,
+            'event_slug': self.event_slug,
             'event_location': self.event_location,
             'event_name': self.event_name,
             'event_organizers_text': self.event_organizers_text,
--- a/civictechprojects/models.py
+++ b/civictechprojects/models.py
@@ -453,7 +453,10 @@ class Event(Archived):
 
     @staticmethod
     def get_by_slug(slug):
-        return Event.objects.filter(event_slug=slug).first()
+        if slug is not None:
+            _slug = slug.strip()
+            if len(_slug) > 0:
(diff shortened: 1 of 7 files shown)


civictechprojects/models.py
```diff
@@ -453,7 +453,10 @@ def hydrate_to_list_json(self):
 
     @staticmethod
     def get_by_slug(slug):
-        return Event.objects.filter(event_slug=slug).first()
+        if slug is not None:
+            _slug = slug.strip()
```
PeterBreen, line 457: I tested my question I asked in slack just by manipulating the querystring directly, and these caused errors in console and a failure to render:
`id=5&dsfgh`
`id=Sponsor&Sponsor` 
and similar uses of ? and &.

this is different from a numerical ID which doesn't exist or a string without special characters, like `id=65` or `id=fakeslug` - both of which correctly render the page and output the message "Could not load Event" as expected.

As slug creation is (I believe?) admin side only right now we can just make sure not to create bad slugs in the short term, but eventually an automatic system (lowercase only replacing all special characters with dashes is the most common pattern I've seen) would help prevent errors here.

PeterBreen (changes requested):
I'm getting an error attempting to create an event. When I put a date into the datepicker, my console coughs up this:
> Starting with v2.0.0-beta.1 date-fns doesn't accept strings as arguments. Please use `parseISO` to parse strings. See: https://git.io/fjule

Ignoring that and attempting to submit / go to page 2 of event creation just hangs on submit. Curiously, the events seem to be created anyway, as viewing My Events shows them:

Screen Shot 2020-11-17 at 6 55 01 PM

Either way, Create Event seems to be having some issues.

PeterBreen (dismissed):

PeterBreen (approved):

PR #456, opened 2020-11-19
Allow camera and microphone in Live Event iframe

marlonkeating:
Qiqochat has released a new tool that requires camera and microphone access for the iframe it lives in.


PeterBreen (dismissed):
image

Well, that was easy to review! 👍

PeterBreen (approved):
Cannot reproduce that login error I had in Chrome/Windows in any other browser/os combination. May be a privacy extension at work? Either way, I feel pretty confident it is not the fault of this PR.

PR #457, opened 2020-11-25
Event Tweaks and Optimization

marlonkeating:
## Caching ##
- Hydrated event data is now cached in the database in the same manner as for projects
- Cached event refreshed when linked projects are edited
- manage.py cache_initialization will also initialize event cache

## Open Graph Support ##
- Open graph name, short description, and image thumbnails are presented for events

## Slug Management ##
- When event slugs are changed, the old slug is saved and the event can still be navigated to with the old slug


civictechprojects/forms.py
```diff
@@ -114,35 +114,45 @@ def create_or_edit_event(request, event_id):
...
             if slug_event and slug_event.id != event.id:
                 print('Could not change event {event} slug to {slug} because another event already has that slug: {existing_event}'.format(
                     event=event.__str__(),
                     slug=slug,
-                    existing_event=slug_event
+                    existing_event=slug_event.__str__()
                 ))
             else:
-                project_fields_changed |= read_form_field_string(event, form, 'event_slug', lambda str: str.lower())
+                # TODO: If we change back to an old slug, delete that NameRecord
```
PeterBreen, line 146: I'm not sure about this TODO. Here's my thinking. An event has slug A and is published. Later that's changed to slug B. Both are 'published' public slugs at some point. We retain knowledge of A even though the canonical slug is B so no links break. Even if the canonical slug is later changed back to A, wouldn't the same logic apply on retaining B?

It's a todo so obviously not blocking, just want to know what the reasoning is.

marlonkeating: Actually this did get done.  You're probably right in that retaining redundant slug records probably wouldn't hurt, but in that case I'd like us to stay consistent, in that we just make a Name Record every time a slug is set, and never delete them.  Does that sound good to you?

PeterBreen: Yeah, that makes a lot of sense to me. Since the business case was 'links don't break' it seems simpler to just keep name records as slugs are set. 👍

PeterBreen (dismissed):
Testing project id=14 via my `slug-one` and `slug-two` slugs shows this works as intended and expected. Nice!

e: I tested OpenGraph through Discord (hey, it was quick...) and I get event title, image, and short description as expected.

PeterBreen (approved):

PeterBreen:
Okay, now that I'm testing it properly, it works fine. 

There's one small thing we may want to fix later, but `Event Url Slug` will always show the first slug created, and not the "current" one.

for example I created an event with slug-one, then changed to slug-two, slug-three, and slug-four in that order. At each step the event url slug field always read "slug-one"

but it's admin only for now so, whatever!

PeterBreen (dismissed):

PR #458, opened 2020-11-25
Don't use project_thumbnail for full-text search

marlonkeating:
We are currently searching the contents of the project_thumbnail file entry in projects, and this has caused some search issues (most significantly, trying to search for 'DemocracyLab', which is present in the s3 file url and so causes nearly every project to come up for the query).

### DIFF
--- a/civictechprojects/models.py
+++ b/civictechprojects/models.py
@@ -219,6 +219,8 @@ class Project(Archived):
         del base_json['project_events']
         del base_json['project_groups']
         del base_json['project_commits']
+        # Don't cache files because they contain noise without adequate signal
+        del base_json['project_thumbnail']
         full_text = str(base_json)
         if len(full_text) >= Project._full_text_capacity:
             full_text = full_text[:Project._full_text_capacity - 1]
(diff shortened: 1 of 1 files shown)


civictechprojects/models.py
```diff
@@ -215,10 +215,17 @@ def recache(self):
...
+        # Don't cache files because they contain noise without adequate signal
+        omit_fields += ['project_thumbnail', 'project_files']
+        # Don't cache boolean fields
+        omit_fields += ['project_claimed', 'project_approved']
+        # Don't cache numeric fields
+        omit_fields += ['project_id', 'project_creator', 'project_latitude', 'project_longitude']
+        # Don't cache date fields
+        omit_fields += ['project_date_modified']
+        for field in omit_fields:
+            base_json.pop(field, None)
```
PeterBreen, line 228: Are there performance concerns with python's pop() here, potentially down the line on a larger set of data to work with? 

It clearly works - I searched for the string democracylab on two environments and got 117 results (including incorrect results) or 6 (correctly) and if I understand how this caching function works it's not likely to be impacting the user experience except in significantly better full-text search results for a few key phrases.

PeterBreen (approved):
One question, not impacting approval.

civictechprojects/models.py
marlonkeating: I'm not overly concerned with performance in this bit of code, as it only runs when objects are updated, as part of an operation that includes a number of database reads.

PR #459, opened 2020-11-26
Donate Page: Show blurb for Giving Tuesday

marlonkeating:
## Donate Page ##
- Show blurb describing Giving Tuesday
- Show blurb by setting environment variable DONATE_PAGE_BLURB='GivingTuesday'
- Preload open graph data

### DIFF
--- a/common/components/componentsBySection/Donate/DonateBlurb.jsx
+++ b/common/components/componentsBySection/Donate/DonateBlurb.jsx
@@ -1,0 +1,46 @@
+// @flow
+
+import React from 'react';
+
+type State = {|
+  blurbType: ?string
+|};
+
+class DonateBlurb extends React.Component<{||}, State> {
+  constructor(): void {
+    super();
+    this.blurbs = {
+      "GivingTuesday": this._givingTuesday
+    };
+    this.state = {
+      blurbType: "GivingTuesday"
+    };
+  }
+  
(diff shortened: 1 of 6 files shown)


common/components/componentsBySection/Donate/DonateBlurb.jsx
```diff
@@ -0,0 +1,47 @@
...
+      </React.Fragment>
+    );
+  }
+  
+  _givingTuesday(): React$Node {
+    return (
+      <React.Fragment>
+        <h1>
+          Giving Tuesday 2020 is on Dec. 1st
+        </h1>
```
PeterBreen, line 34:
```suggestion
        <h2>
          Giving Tuesday 2020 is on Dec. 1st
        </h2>
```

In the short term this makes the title text a little smaller but also has the H1/H2/H3 ... hierarchy.

PeterBreen (approved):
Seems to work, and I like using the env key to just point at a predefined blurb of text rather than trying to contain the whole thing like the unwieldy PRESS_LINKS style.

Let's talk Monday if you want me to do a micro-fix to clean up a few more rough edges on this page's style prior to December 1. I can probably get something put together in half an hour. It'd be a low-risk hotfix but it would be a hotfix so I understand if that's not appealing.

PR #499, opened 2020-12-01
Project Profile: Don't hide actions for embedded view

marlonkeating:


marlonkeating:
These changes will be included as part of #521

PR #500, opened 2020-12-01
Add sitemap entries for Groups and Events

marlonkeating:
- Add sitemap entries for all approved Groups
- Add sitemap entries for all approved, non-private Events
- Add unit tests for front_end.py


PeterBreen (approved):
Looking good to me. I checked the XML and it looks like it's got the right groups/events listed.

PR #502, opened 2020-12-02
Peter/reply volunteer email

PeterBreen:
Updates the email sent to a project owner when a new volunteer submits the volunteer form on DemocracyLab.

### to do:
 - [x] if statement on Zip Code line (do not display zip line at all if user.postal_code is empty)

(Partial) Current output example:
Screen Shot 2020-12-04 at 11 27 26 AM

### DIFF
--- a/democracylab/emails.py
+++ b/democracylab/emails.py
@@ -15,6 +15,8 @@ from common.helpers.front_end import section_url
 
 class EmailSection(Enum):
     Header = 'Header'
+    Headerleft = 'Headerleft'
+    Subheader = 'Subheader'
     Button = 'Button'
     Paragraph = 'Paragraph'
     Paragraph_Center = 'Paragraph_Center'
@@ -30,6 +32,8 @@ class HtmlEmailTemplate:
     base_template = loader.get_template('html_email_frame.html')
     section_templates = {
         EmailSection.Header: loader.get_template('html_email_header.html'),
+        EmailSection.Headerleft: loader.get_template('html_email_headerleft.html'),
+        EmailSection.Subheader: loader.get_template('html_email_subheader.html'),
         EmailSection.Button: loader.get_template('html_email_button.html'),
         EmailSection.Paragraph: loader.get_template('html_email_paragraph.html'),
         EmailSection.Paragraph_Center: loader.get_template('html_email_paragraph_center.html')
@@ -205,7 +209,15 @@ def send_volunteer_application_email(volunteer_relation, is_reminder=False):
         project=project.project_name,
         role=role_text)
     email_template = HtmlEmailTemplate()\
-        .header("You Have a New Volunteer!")\
+        .subheader("Opportunity Information:")\
+        .paragraph("Title:")\
+        .paragraph("Organization:")\
+        .paragraph("Date:")\
+        .subheader("Volunteer Information:")\
+        .paragraph("Name:")\
+        .paragraph("Email:")\
+        .paragraph("Phone:")\
+        .headerleft("You Have a New Volunteer!")\
         .paragraph('\"{message}\" -{firstname} {lastname}'.format(
             message=volunteer_relation.application_text,
             firstname=user.first_name,
--- a/democracylab/emails.py
+++ b/democracylab/emails.py
@@ -20,6 +20,7 @@ class EmailSection(Enum):
     Button = 'Button'
     Paragraph = 'Paragraph'
     Paragraph_Center = 'Paragraph_Center'
+    Text_Line = 'Text_Line'
 
 
 class Html:
@@ -36,7 +37,8 @@ class HtmlEmailTemplate:
         EmailSection.Subheader: loader.get_template('html_email_subheader.html'),
         EmailSection.Button: loader.get_template('html_email_button.html'),
         EmailSection.Paragraph: loader.get_template('html_email_paragraph.html'),
-        EmailSection.Paragraph_Center: loader.get_template('html_email_paragraph_center.html')
+        EmailSection.Paragraph_Center: loader.get_template('html_email_paragraph_center.html'),
+        EmailSection.Text_Line: loader.get_template('html_email_text_line.html'),
     }
 
     def __init__(self, use_signature=True):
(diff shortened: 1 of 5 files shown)


democracylab/emails.py
```diff
@@ -205,12 +220,23 @@ def send_volunteer_application_email(volunteer_relation, is_reminder=False):
         project=project.project_name,
         role=role_text)
     email_template = HtmlEmailTemplate()\
-        .header("You Have a New Volunteer!")\
+        .subheader("Opportunity Information:")\
+        .text_line("Title: {role}".format(role=role_details.display_name))\
+        .text_line("Organization: {project}".format(project=project.project_name))\
+        .text_line("Date: {currentdate}".format(currentdate=models.DateTimeField(auto_now_add=True, null=True)))\
```
marlonkeating, line 226: models.DateTimeField is specifically for declaring a database column.  If you want to get the current date/time, use timezone.now() and pass it into date_helpers.datetime_to_string.

Example: https://github.com/DemocracyLab/CivicTechExchange/blob/fa6633d16eadc0cdacea7c052b68fdf6b1fc8e98/democracylab/emails.py#L253

democracylab/emails.py
```diff
@@ -30,9 +33,12 @@ class HtmlEmailTemplate:
     base_template = loader.get_template('html_email_frame.html')
     section_templates = {
         EmailSection.Header: loader.get_template('html_email_header.html'),
+        EmailSection.Headerleft: loader.get_template('html_email_headerleft.html'),
+        EmailSection.Subheader: loader.get_template('html_email_subheader.html'),
         EmailSection.Button: loader.get_template('html_email_button.html'),
         EmailSection.Paragraph: loader.get_template('html_email_paragraph.html'),
-        EmailSection.Paragraph_Center: loader.get_template('html_email_paragraph_center.html')
+        EmailSection.Paragraph_Center: loader.get_template('html_email_paragraph_center.html'),
+        EmailSection.Text_Line: loader.get_template('html_email_text_line.html'),
```
marlonkeating, line 41:
```suggestion
        EmailSection.Text_Line: loader.get_template('html_email_text_line.html')
```

democracylab/emails.py
```diff
@@ -15,9 +15,12 @@
 
 class EmailSection(Enum):
     Header = 'Header'
+    Headerleft = 'Headerleft'
```
marlonkeating, line 18: nit: Let's try to keep the casing consistent 
```suggestion
    Header_Left= 'Header_Left'
```

democracylab/emails.py
PeterBreen: and this is why I need to figure out how to get linting working on .py files. Thanks, good eye.

democracylab/emails.py
PeterBreen: Got it. I ... hm, I must not have committed a change I made where I copied from a better example in emails.py but it wasn't timezone.now() so I'll fix this, thanks.

democracylab/emails.py
```diff
@@ -222,20 +223,19 @@ def send_volunteer_application_email(volunteer_relation, is_reminder=False):
...
-        .text_line("Organization: {project}".format(project=project.project_name))\
-        .text_line("Date: {currentdate}".format(currentdate=models.DateTimeField(auto_now_add=True, null=True)))\
+        .text_line("Organization: {projectname}".format(projectname=project.project_name))\
+        .text_line("Date: {currentdate}".format(currentdate=datetime_to_string(timezone.now(), DateTimeFormats.DATE_LOCALIZED)))\
         .subheader("Volunteer Information:")\
         .text_line("Name: {firstname} {lastname}".format(
             firstname=user.first_name,
             lastname=user.last_name))\
-        .text_line("Email: {email}".format(email=Html.a(href='mailto:' + user.email, text=user.email)))\
+        .text_line("Email: {email}".format(email=Html.a(href='mailto:{useremail}'.format(useremail=user.email), text=user.email)))\
```
PeterBreen, line 232: The href output here is missing quotes, per my dev console. That is, I get this:
`<a href=mailto:email@address.com>email@address.com</a>`

instead of what I want, which is this:
`<a href="mailto:email@address.com">email@address.com</a>`

Also it feels like nested .format() is pretty messy, is there a better way to do it?

democracylab/emails.py
```diff
@@ -222,20 +223,19 @@ def send_volunteer_application_email(volunteer_relation, is_reminder=False):
...
         .text_line("Zip: {zip}".format(zip=user.postal_code))\
```
PeterBreen, line 233: There's no conditional formatting here (and I'm not really sure why we ask for this field / put it in the email, since it's not like I've memorized every Seattle zip code to know if a volunteer is "local" or not?) -- is the `is_reminder` if statement a good pattern to follow for conditionally showing or hiding this? I think having a blank Zip: field is pretty awkward.

democracylab/emails.py
```diff
@@ -205,12 +221,22 @@ def send_volunteer_application_email(volunteer_relation, is_reminder=False):
         project=project.project_name,
         role=role_text)
     email_template = HtmlEmailTemplate()\
-        .header("You Have a New Volunteer!")\
+        .subheader("Opportunity Information:")\
+        .text_line("Title: {role}".format(role=role_details.display_name))\
+        .text_line("Organization: {projectname}".format(projectname=project.project_name))\
+        .text_line("Date: {currentdate}".format(currentdate=datetime_to_string(timezone.now(), DateTimeFormats.DATE_LOCALIZED)))\
```
PeterBreen, line 227: The date format this outputs is mm/dd/yy, when the design request has Month DD, YYYY. Should that be `%B %d %Y` in DateTimeFormats?

democracylab/emails.py
marlonkeating: We'll want to format things a bit differently here.  Essentially, we don't want to include the line at all if there's no zip, so we can do something like

email_template = HtmlEmailTemplate()\
// all the lines up till the zip line
if user.postal_code:
  email_template = email_template.text_line("Zip: {zip}".format(zip=user.postal_code))
// everything else

democracylab/emails.py
marlonkeating: There's an issue with the Html.a helper.  It's not including quotes for the href part.

Also, for simple string appends like this you can use addition instead.
```suggestion
        .text_line("Email: " + Html.a(href='mailto:' + user.email, text=user.email))\
```

democracylab/emails.py
marlonkeating: You can create another Enum with the appropriate format string (yours looks correct to me, just don't forget the comma!)

democracylab/emails.py
PeterBreen: Made that change, and I think I made Html.a include quotes the way we want on top. Triple quotes apparently do it?

democracylab/emails.py
```diff
@@ -27,7 +27,7 @@ class EmailSection(Enum):
 class Html:
     @staticmethod
     def a(href, text):
-        return "<a href={href}>{text}</a>".format(href=href, text=text)
+        return """<a href="{href}">{text}</a>""".format(href=href, text=text)
```
PeterBreen, line 30: Wanted to call this out - I saw several methods to get our href output to include quotes, and I went with this since it was fast and human legible. If another method looks cleaner now's as good a time as any to try it since testing will have to go beyond just the volunteer email since Href.a is used elsewhere.

marlonkeating: Ordinarily we use single quotes for strings that will include double quotes, and vice versa.  I went ahead and changed this line to that format in my latest change.

marlonkeating (approved):

PR #503, opened 2020-12-04
Event Profile: Fix private event error 403 for non-admin/creator

marlonkeating:
Users who are authenticated but not the creator or an admin currently get an error 403 when viewing a private event using the slug id.  The error 403 should only happen when the user has navigated to the private event using its numeric id.


PeterBreen (approved):
Logged out via number: 403
Logged in via number: 403

Logged out via slug: loads correctly
Logged in via slug: loads correctly

PR #504, opened 2020-12-05
Require project creators to agree to terms of service

marlonkeating:
# Overview # 
Besides users agreeing to terms of service (that primarily governs their activities as a volunteer), we want project owners to agree to  certain terms as well.

## Create Project Page ##
- Require creator to check box signifying they agree with terms of service
- Provide link to modal that shows organization terms of service


PeterBreen (approved):
Looks good. I checked the volunteer signup and the create project modals and they both display the correct text for the correct modal.

PR #505, opened 2020-12-07
OAuth Thumbnail Fixes

marlonkeating:
# Background #
Currently when a user signs in via OAuth, we set as their thumbnail the file provided to us by the OAuth provider.  However, the file path given to us often does not allow it to be displayed on our site, or if it does it's only for a limited time.  Therefore, we want to be able to save the image to our S3 for persistent storage.

## Changes ##
- Now when a user logs in via OAuth, their thumbnail will be saved to our S3 immediately, and referenced thereafter.

## Backfilling ##
- To upload all existing OAuth user thumbnails (that are still available for us to download), run the command: python manage.py data_migration upload_oauth_thumbnails
- Any thumbnails that are no longer available will be marked as type 'THUMBNAIL_ERROR', and we will no longer attempt to show it on the front end

## Admin Portal ##
- Added ProjectFile support


PeterBreen (approved):
Looks good. I went through the deployed version as well as the code and I see only user images or the correct default placeholder image; no more awkward empty spaces.

PR #515, opened 2020-12-21
Library upgrades to support Heroku-18 stack

marlonkeating:
Our current heroku stack is being phased out in a few months, and so we're updating our libraries to work with Ubuntu 18, which is what the new heroku stack uses.

## Library Updates ##
-Update package.json node version to 14.x, which is the default node version for Ubuntu 18
-Updated babel to fix build issues with node 14
-Updated flow to fix intellisense issues with other updated libraries
-Updated flux to fix errors


PeterBreen:
Just a few things I've come across by smoke testing. I'll add to this as I continue.

#### sitewide

 - the get header height function that sets padding to push our page content below the fixed navbar isn't working, so the tops of every page are getting clipped off. I believe that's MainHeader.jsx and/or MainController.jsx?


#### find projects, find groups

 - all filters except Location do not load 
 
 
#### logging in

 - after entering username/pw and clicking login, page fails to render, gives a somewhat-not-useful error "e is not defined." The actual act of logging in appears to work though, because on refresh/reload I see a render for a frame or two which includes my user information in the top right corner, but then it goes blank and I get 'e is not defined' in console again.

PeterBreen (approved):
Passes smoke testing right now, looks good to merge into the library changes I did earlier prior to a full regression test.

🎉

Issue #516, opened 2020-12-30
Edit Project: Slow performance navigating between pages


marlonkeating:
When editing a project, it consistently takes 5 seconds or more to move to the next page without making any changes (via the Next button), when it should be nearly instantaneous.

PR #517, opened 2020-12-30
Django 3 upgrade

marlonkeating:
Upgrading from django 1.11 to 3.1 in order to take advantage of new django features.


PeterBreen (approved):
Smoke tested this from the browser side and it looks fine, read through some code changes and that also looks fine. Specifically, page navigation, sign-in, editing a project, contacting a project, contacting DemocracyLab, and rendering all DL pages had no user-facing errors or render failure.

As a potential side bonus, #516 may be resolved here? I replicated the steps in that issue on this branch and saw page loads take no longer than about 1s. I didn't time it exactly but there's no chance they were 5s or greater.

PR #518, opened 2021-01-05
Don't show un-approved group/project relationships

marlonkeating:
Currently, if a group/project relationship is un-approved (either hasn't been approved or the approval has been revoked), the group profile still shows the project listed, and the project profile still lists the group.


PeterBreen (changes requested):
Generally this works but I found a case where it seems not to. On dev, I created a group and a project and then associated the two. As the owner of both it was automatically approved.

Both group and project are unapproved, but the project profile shows the group relation. The group page does not show the project relation. This doesn't seem to be intended behavior?

https://democracy-lab-dev.herokuapp.com/index/?section=AboutProject&id=588
https://democracy-lab-dev.herokuapp.com/index/?section=AboutGroup&id=14

PeterBreen (approved):

PR #521, opened 2021-01-09
Support embedded project actions

marlonkeating:
We want to be able to perform project actions like contacting and volunteering with projects within the embedded view used by qiqochat.

## Project Profile ##
- No longer hide actions within embedded view

## Cookies ##
- Session and CSRF cookies are now set to SameSite=None to support being loaded within cross-domain iframes
- Session and CSRF cookies are now marked Secure for non-debug environments

## Misc Changes ##
- gitignore VS Code settings file


PeterBreen (dismissed):
Looks pretty good in the prod-mirror test. For release test cases let's make sure an event that isn't DemocracyLab (i.e. an event where I'm not already a member of the team) is part of it, just in case?

PeterBreen (approved):
CSP looks good and the embedded actions were fine as well.

Issue #522, opened 2021-01-11
Find Projects: Use cached project information to render project cards


marlonkeating:
We are experiencing performance issues on the Find Project page.  As we are observing that the page load time is strongly dependent on how many cards are returned by the search query, we should optimize the card rendering.

Currently each project card does at least one additional join for their thumbnail, so if we can leverage the already-cached project json (which includes the thumbnail) we should be able to improve the rendering performance.

PR #524, opened 2021-01-12
 Used cached project data to render project cards

marlonkeating:
To fix performance issues on the Find Project page. we now leverage the already-cached project json (which includes the thumbnail), and this greatly improves the rendering performance.

Addresses #522

## Other Changes ##
- Added unit tests for dictionary helpers


PeterBreen (approved):
This is definitely helping with long response times from the db; overall rendering feels quicker (and is according to some tests I run) and while there's still some inconsistency in overall response times, it's inconsistently just a bit faster 👍

PR #525, opened 2021-01-14
Event project search

marlonkeating:
# Overview # 
This change adds project filtering to the Event Profile page

## Performance Optimizations ##
- Find Events: Use cached event data to render tiles
- Event Profile: Don't return event list as part of event profile payload
- Create Project: Don't recache project if no fields were changes

## Misc Changes ##
- Add prettier configuration
- Exclude node and git folders, VS code config from git


.vscode/settings.json
```diff
@@ -1,3 +1,8 @@
 {
-    "python.linting.pylintEnabled": false
+    "python.linting.pylintEnabled": false,
+    "editor.fontSize": 16,
+    "files.exclude": {
```
PeterBreen, line 4: Oh, this took a font size change I didn't want it to. What do you think about adding it to gitignore instead? Or maybe get it in a good 'default' state and then gitignoring it so further changes don't get passed up?

marlonkeating: Yeah, let's gitignore it for now

PeterBreen: Well that took a few more commits than I had planned but 248ea25 ought to do it. Since git 'knew' about .vscode/settings.json just gitignoring it wasn't enough, so I followed the instructions [from here](https://stackoverflow.com/questions/11451535/gitignore-is-ignored-by-git)
```
git rm -r --cached .
git add .
git commit -m "fixed untracked files"
```

It worked - but it also removed these, so I wonder if we had a few other orphaned files kicking around. Nothing seems to have broken after they were removed, at least.
```
Changes to be committed:
  (use "git reset HEAD <file>..." to unstage)

	deleted:    .vscode/settings.json
	deleted:    common/components/common/tags/TagCategory.jsx
	deleted:    common/components/common/tags/TagSelectWrapper.jsx
	deleted:    common/components/common/tags/TagSelector.jsx
	deleted:    common/components/common/tags/TagsDisplay.jsx
```

PeterBreen (approved):
I've reviewed all the changes in this PR that you did and it all looks good. Also in terms of the user experience it all works, a nice bonus.

PR #532, opened 2021-01-26
Prettier codemod

PeterBreen:
- Create `.prettierignore` config file
 - Run prettier on our js/css/html/md files
 - Generate world's largest diff

Okay maybe not the last one. Should involve no code changes, only prettier formatting.


civictechprojects/templates/new_index.html
```diff
@@ -1,26 +1,29 @@
...
+    <link rel="stylesheet" type="text/css" href="/static/css/main.styles.css" />
+    <link rel="icon" type="image/png" href="{{FAVICON_PATH}}" />
+    <link
+      rel="stylesheet"
+      href="https://use.fontawesome.com/releases/v5.5.0/css/all.css"
+      integrity="sha384-B4dIYHKNBt8Bc12p+WXckhzcICo0wtJAoU8YZTY5qE0Id1GSseTk6S+L3BlXeVIU"
+      crossorigin="anonymous"
+    />
+    {{ googleTagsHeadScript }} {{ organizationSnippet }} {{ hotjarScript }} {{
+    googleScript }}
```
PeterBreen, line 26: Seeing some problems with this section in Firefox and Chrome, where `{{ googleScript }}` is rendered literally. We'll likely want to set this block to be ignored or to not use Prettier on our HTML, depending.

image

.prettierignore
```diff
@@ -0,0 +1,46 @@
...
+!*.json
+
+
+# css/sass
+!*.scss
+!*.css
+
+# html
+!*.html
+!*.htm
```
marlonkeating, line 25: It looks like prettier doesn't know how to handle embedded django template directives, so let's omit html files for now.

image

marlonkeating (changes requested):

.prettierignore
PeterBreen: Should be resolved now. I updated the prettierignore and then reverted all HTML file changes to current master, 053d576.

marlonkeating (approved):

PR #533, opened 2021-01-27
Group project search

marlonkeating:
# Group Project Search #
This change adds project filtering to the Group Profile page

## Performance Optimizations ##
- Group Profile: Use cached group data
- Find Groups: Use cached group data to render tiles


PeterBreen (approved):
Checked over all the code and it looks fine. I appreciate the toggle change for expand/collapse not being just the arrow and instead being the whole filter box - saw that get in there. And the perf improvements are excellent.

Assuming the change I made to the ProfileProjectSearch component pass muster, approved.

PR #537, opened 2021-02-04
Heap analytics

marlonkeating:
- Add Heap analytics blurb to main page header
- Environment variable HEAP_ANALYTICS_ID for controlling which heap environment to use
- Update sponsors to include Heap


PeterBreen (commented):
Sponsors look good, but I don't see the heap analytics loading. Do we have a test environment ID we can use to check that without adding extraneous pageloads to the analytics?

marlonkeating:
We do have a test heap account, which I've just enabled on the dev environment.

PeterBreen (approved):

PR #543, opened 2021-02-23
Update url routing

marlonkeating:
This change institutes a new url scheme across the site, to follow a more standard format.

## Changes ##
- Old urls will redirect to their new urls
- Back-end urls that clashed with new front-end urls are now prefixed with /api


PeterBreen (approved):
After the full test pass and then these fixes I think we're good on our routing.

PR #544, opened 2021-02-25
Event Profile: Enable clickable links in Details/Agenda

marlonkeating:
We want to let event creators add links to their details/agenda, and make those links clickable.  We are using the react-linkify library to enable this.


PeterBreen (approved):
Looks straightforward; we discussed some potential pitfalls in standup but I think it's fine, especially as a starting point to develop better information for potential expansion later.

PR #546, opened 2021-03-02
Fix re-caching groups search tags

marlonkeating:
When navigating to the group profile page, the cached tags for all projects get overwritten by the subset of tags used for that group's projects.  This change fixes that.


PeterBreen (approved):
Works great 👍

PR #548, opened 2021-03-03
Docker environment fixes

marlonkeating:
We've recently been seeing issues spinning up a new development environment database due to errors in geospatial field migrations.  This change fixes the migrations to put in a valid blank location string.

## Other Changes ##
- Fixed backend unit tests to reflect new v2 urls


PeterBreen (approved):
This has now been confirmed to work for two Docker setups, one brand new and one not, so it's good to go in my view.

PR #552, opened 2021-03-09
Event Profile: Sort projects by name by default

marlonkeating:
On the Event profile page, we'd like the projects to be sorted in alphabetical order by default rather than last updated date, in order to give a stable view of the project list when presenting at these events.


PeterBreen (approved):
Looks good, checked it against live and it seems to work just fine.

PR #554, opened 2021-03-10
Project create/profile fixes

PeterBreen:
Fixes for #549, fixes to Create Project page, also some release testing bugs

### DIFF
--- a/civictechprojects/models.py
+++ b/civictechprojects/models.py
@@ -152,7 +152,7 @@ class Project(Archived):
             'project_date_modified': self.project_date_modified.__str__()
         }
 
-        if self.project_location_coords is not None:
+        if self.project_location_coords is not None and not self.project_location_coords.empty:
             project['project_latitude'] = self.project_location_coords.x
             project['project_longitude'] = self.project_location_coords.y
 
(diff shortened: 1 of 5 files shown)


PeterBreen:
some bugs fixed, for documentation

- [x] logo min-size rules need revision so it doesn't break awkwardly in tablet
- [x] get small logos to take up space properly, probably remove flex-basis @ desktop, set logo img: width 100%; set container width manually (250px?) - potentially display flex on parent div with align-items: flex-start; -- see /projects/705 on live vs prod-mirror for what's wrong
- [x] long text fields break in details tab, skills needed tab, likely elsewhere (set overflow-wrap and/or word-break)
- [x] long URL in top details section breaks the whole thing, constrain maximum width, possibly need above rules here too
- [x] long names in secondary column, including Files, Team section push that entire column to be wider rather than breaking onto 2 lines (names) or truncating (files) as appropriate - easiest fix is probably fixed width above lg breakpoint
- [x] Team list, `PersonIcon` SVG should be raised to 70x70 so the container can be 80x81, matching an uploaded VolunteerCard image
- [x] rethink use of `overflow: hidden; text-overflow: ellipsis;` ellipsis over `word-break: break-word` for project files section. see projects/646/ for a good example for usability testing
- [x] fix `<a>` font size on mobile

PR #565, opened 2021-03-17
Fix back-end url generation

### LINKED ISSUE #564: Reset Password Fails due to malformed url
When a user clicks Forgot Password and enters their email address, they are sent an email with a malformed url to the Reset Password page, that fails to reset the password and leads to an error page.

marlonkeating:
During the recent url revision ( #543 ), back-end generation of urls that have multiple arguments broke, resulting in urls that do not include the leading ? to the argument string.  

fixes #564

### DIFF
--- a/civictechprojects/tests/test_views.py
+++ b/civictechprojects/tests/test_views.py
@@ -1,8 +1,9 @@
+from unittest import skip
 from django.test import TestCase, Client, tag
 from django.urls import reverse
 from civictechprojects.models import Project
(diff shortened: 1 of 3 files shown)


PeterBreen (dismissed):

PeterBreen (approved):
Tests pass, looks good.

PeterBreen (approved):
Tests pass, URLs work, broken tests are skipped.

PR #566, opened 2021-03-18
Allow code to be hot-reloaded with Docker

KentShikama:
We now mount in the source code instead of copying it into the image: this allows us to propagate changes in our host file system to the running Docker container.

Test as follows:

```
$ docker-compose down -v
$ docker-compose up --build
```

And then try changing a random api url path while the containing is running. It should reflect immediately upon hard refresh.

Worked with @glenbraun @marlonkeating

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,8 +31,6 @@ RUN yarn config set ignore-engines true
 RUN yarn --frozen-lockfile --link-duplicates --ignore-scripts
 # Permission issue with node-sass https://github.com/sass/node-sass/issues/1579
 RUN npm rebuild node-sass
-# Copy folders and files whitelisted by .dockerignore.
-COPY . /code/
 
 # Password for account used to send email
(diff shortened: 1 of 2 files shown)


KentShikama:
I ran the following steps on a different machine to what I originally developed this PR with

```
$ git clone git@github.com:DemocracyLab/CivicTechExchange.git
$ cd CivicTechExchange/
$ docker --version
Docker version 20.10.5, build 55c4c88
$ docker-compose --version
docker-compose version 1.28.5, build c4eb3a1f
$ git remote add kent git@github.com:KentShikama/CivicTechExchange.git
$ git fetch kent
$ git checkout kent/kent/docker
$ docker-compose down -v
$ docker-compose up --build
```

Here is a dump of the output of the last command: https://gist.github.com/KentShikama/9203e0ab95e59ccc630afb15d26b5697.

marlonkeating:
Changes have been incorporated as part of #626

PR #569, opened 2021-03-19
Add meta image property to mirror og:image property

### LINKED ISSUE #568: Add meta name="image" tag to context preload
Currently some platforms that support Open Graph do not support the meta name="og:image" tags we publish, but use meta name="image" instead.  We should publish both tags.

ex: `<meta name="image" property="og:image" content="[Image URL here]">`

## Testing ##

- [ ] https://www.linkedin.com/post-inspector/

marlonkeating:
Some OpenGraph providers read the 'image' meta tag rather than the 'og:image' tag, so this change adds that.

closes #568


PeterBreen:
Looks good on LinkedIn's post inspector (as specified in #568) and doesn't regress in any of my other OG testers. 👍

PeterBreen (approved):

PR #573, opened 2021-03-22
Group profile page update

PeterBreen:
- Updates group profile page to use the same template as project profile page
 - Adds LoadingFrame page load spinner
 - removes old MUI-based CSS, cleans up new AboutProject/AboutGroup CSS to remove duplication
 - fixes React unique key warnings on skills needed/tech used sections


PeterBreen:
Issue tracking before this is ready for review:
- [ ] review of CSS classnames so we have reusable Profile- and unique AboutGroup-/AboutProject-
- [ ] white-space: pre-wrap; on main pane content for about group (possibly some other rules because we're not using tabs)
- [ ] fix whatever is going on [here](https://democracy-lab-dev.herokuapp.com/groups/12) and [here](https://democracy-lab-dev.herokuapp.com/groups/7)
     - [ ] possibly rework the card container component; it's doing things we may not want carried over here
- [ ] a non-placeholder title for group description, rendered conditionally - if group description isn't required (?)

common/components/common/groups/AboutGroupDisplay.jsx
```diff
@@ -68,7 +69,7 @@ class AboutGroupDisplay extends React.Component<Props, State> {
     return this.state.group ? (
       this._renderPageLayout()
     ) : (
-      <div>{this.state.loadStatusMsg}</div>
+      <LoadingFrame height={600} />
```
PeterBreen, line 72: I think this conditional check may be irrelevant because the AboutGroupController also has conditional checks - that is, `this.state.group` will always be true because this entire component isn't rendered until the group data has returned from the API. The same is true of AboutProjectController / AboutProjectDisplay. See the changes in commit 1d01dd5. 

If that's the case I can just remove this conditional here and in AboutProjectDisplay, clean those up a bit.

It may be worth the time (possibly in another PR) to refactor out AboutProjectController and AboutGroupController to remove helmet, produce error messages in another fashion, and

marlonkeating: Yeah, this check is done by the parent in all the current cases.

marlonkeating (changes requested):
Overall looks good, just a couple minor cleanups I can see.

- Remove Headers from AboutGroupController

marlonkeating (approved):

PR #574, opened 2021-03-24
Users supplies missing first/last name when signing up via OAuth

marlonkeating:
When users sign up using an OAuth provider that doesn't require users to provide their full names (such as Github), the signup currently fails if their full name was not set.  This change introduces a new page that lets users in this scenario enter their first and last names before they can proceed on browsing the site.

### DIFF
--- a/civictechprojects/static/css/partials/_LogInController.scss
+++ b/civictechprojects/static/css/partials/_LogInController.scss
@@ -9,6 +9,10 @@
   overflow: auto;
 }
 
+.LogInController-root b {
+  font-weight: bold;
+}
+
 .LogInController-input {
   border: solid 1px $color-grey-frame-border;
   height: 40px;
--- a/civictechprojects/static/css/partials/_LogInController.scss
+++ b/civictechprojects/static/css/partials/_LogInController.scss
@@ -1,16 +1,11 @@
 .LogInController-root {
   background-color: $color-background-light;
   border: solid 1px $color-grey-frame-border;
-  font-weight: 300;
   width: 90%;
   max-width: 800px;
   margin: 76px auto;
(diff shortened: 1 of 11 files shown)


civictechprojects/static/css/partials/_LogInController.scss
```diff
@@ -9,6 +9,10 @@
   overflow: auto;
 }
 
+.LogInController-root b {
```
PeterBreen, line 12: Rather than declare this here, I'd suggest removing this and the `font-weight: 300;` declaration on line 4. That weight is inconsistent with the rest of our text (typically 400) and is the root cause of why bold didn't appear bold before.

I had to look this up to figure it out; [relative font weighting](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#meaning_of_relative_weights) - which we use - doesn't really do what's expected with a base of 300 weight.

common/components/controllers/AddSignUpDetails.jsx
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  render(): React$Node {
+    return (
+      <React.Fragment>
+        <div className="LogInController-root">
+          <div className="LogInController-greeting">
+            <p>
+              We were not able to obtain your <b>First Name</b> and{" "}
+              <b>Last Name</b> from {this.state.service}.
```
PeterBreen, line 54: By preference, I'd use `<strong>` tags rather than `<b>` here.

PeterBreen (changes requested):
I tested OAuth via Github and it prompted me for a first and last as expected, filling those in worked. Aside from the text formatting changes I've noted I don't see any problems.

PeterBreen (approved):

PR #575, opened 2021-03-24
Add in-kind sponsors to footer

marlonkeating:
This change adds sponsors that contribute to DemocracyLab with free or discounted services, and creates a separate section for them (not showing them on the Partner page).


PeterBreen (approved):
Tested this in two environments; sponsors load correctly and the index fix resolves the error page I was seeing on pages with numerical IDs.

Issue #536, opened 2021-02-02
After navigating to Signup Page then Login Page, signing in brings user back to Signup page 


marlonkeating:
When a user clicks on the Log In button on the top header, and then logs in from that page, they are taken back to the page where they clicked the Log In button.  However, this leads to an undesirable scenario when the user goes from the Sign Up page to the Sign In page, and then logs in, resulting in them being taken back to the Sign Up page (which clearly isn't useful if they've successfully signed in).

PR #577, opened 2021-03-30
fix for story 536, route to home page after login

Arpita-22:
I have added the logic to route user from Signup page to home page. 
Story - https://github.com/DemocracyLab/CivicTechExchange/issues/536

### DIFF
--- a/common/components/controllers/LogInController.jsx
+++ b/common/components/controllers/LogInController.jsx
@@ -25,7 +25,7 @@ class LogInController extends React.Component<Props, State> {
     super(props);
     const args: Dictionary<string> = url.arguments();
     let checkPrevPage: string = props.prevPage || args["prev"];
-    const prevPage: string = (checkPrevPage === 'SignUp' ? "Home" : checkPrevPage);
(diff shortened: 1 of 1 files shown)


common/components/controllers/LogInController.jsx
```diff
@@ -24,7 +24,8 @@ class LogInController extends React.Component<Props, State> {
   constructor(props): void {
     super(props);
     const args: Dictionary<string> = url.arguments();
-    const prevPage: string = props.prevPage || args["prev"];
+    let checkPrevPage: string = props.prevPage || args["prev"];
+    const prevPage: string = (checkPrevPage === 'SignUp' ? "Home" : checkPrevPage);
```
marlonkeating, line 28: Use the constants Section.Home and Section.SignUp rather than hard-coding the strings.

marlonkeating (changes requested):

common/components/controllers/LogInController.jsx
Arpita-22: I have made the changes as requested. Please take a look and let me know.

marlonkeating (approved):

PR #584, opened 2021-04-06
corporate page 2021

PeterBreen:
Rebuild of corporate engagement page, with "Host a Hackathon" and "Sponsorship" main content sections.

See comment below for specifics.


PeterBreen:
There are some requested items which didn't make it this PR. They are (with reasoning in _italics_ ):
- refactor testimonial card for mobile (avatar to left of title, not on top) -- _have to confirm this is OK on homepage and here, check in with design_
- update CSP and/or CDN for blog images so they load correctly -- _requires some manual work for old blog posts, so low reward for effort_
- make "at least one" of the three content form checkboxes required - will require JS validation, and must present use case ("at least one or more") and then error if none are checked to user _will handle in a future sprint, since it has UI considerations on presenting errors that should be consistent with the form changes we've got coming_
- contact form checkboxes are bigger/styled as requested (see [figma](https://www.figma.com/file/WADcmVjJh5ARVoZ09xlpfdFN/DemocracyLab?node-id=15291%3A417)) -- _ran out of time; the fast approach wasn't working for all browsers so I've got to go back in later and do it properly_
- hackathon 3-step "how it works" connecting line -- _dropped the ball on scoping this with design; it's more complex than the figma suggested and needs another pass_

Capturing the checklist I worked from for reference:
#### MVP
- [x] headers from backend
- [x] optimized fields returned in Ghost API call (`slug, feature_image, title, custom_excerpt, excerpt, url`)
  - [x] all relevant blog posts are in Ghost
  - [x] blog posts are correctly tagged for us to filter on
  - [x] corporate page requests the correct tags per section
- [x] two testimonials (one per section) - from db
  - [x] new testimonial(s) added
  - [x] correct testimonial data in each section
- [x] anchor link component that uses `<MainController> state.headerHeight` for offset calculation
- [x] carousel autoadvance disabled (it's set to something like 10 minutes because passing null/false didn't take, so it's effectively disabled)
- [x] ability to have links on dlab open the sponsorship tab on initial load rather than Hackathon (which is/should be default unless specifically overridden)
- [x] page style is complete (pretend this is something like 20 checkboxes)
  - [x] hackathon section
    - [x] how it works mobile/desktop
  - [x] sponsorship section
    - [x] image on cdn/in page
    - [x] mobile/desktop layout (flex)
  - [x] shared top section
  - [x] shared contact form
  - [x] shared bottom image section
  - [x] spacing/finishing check, mobile
  - [x] spacing/finishing check, desktop

#### nice-to-haves
- [x] add character limit and character counter to Contact form "Message" field
- [x] loading spinner or something for blog posts, with a min-height set to avoid content shift; reusable for other places on site

common/components/controllers/CorporateHackathonController.jsx
```diff
@@ -1,313 +1,451 @@
...
+          <div className="carousel-blog-root">
+            <BlogCarousel tag="partner-highlights" interval={600000} />
+          </div>
+          <p className="size-h2">
+            Read about other previous events in our{" "}
+            <a className="size-h2" href={window.BLOG_URL} target="_blank">
+              blog
+            </a>
+            , or other events that are coming up{" "}
+            <a className="size-h2" href="/events">
```
marlonkeating, line 272:
```suggestion
            <a className="size-h2" href={url.section(Section.FindEvents)}>
```

common/components/forms/ContactForm.jsx
```diff
@@ -129,88 +135,130 @@ class ContactForm extends React.Component<Props, State> {
...
+                  className="form-control"
+                  name="emailaddr"
+                  type="email"
+                  id="emailaddr"
+                  placeholder="name@example.com"
+                  value={this.state.emailaddr}
+                  onChange={this.handleInputChange}
+                />
+              </div>
+              {this.props.showInterests && this._renderInterestFields()}
```
marlonkeating, line 179: List showInterests in top Props section

marlonkeating (changes requested):
Overall looks good.  Only major issue I saw is the Event Sponsorship checkbox isn't showing up in the contact email that gets sent.

marlonkeating (approved):

PR #596, opened 2021-04-27
Flag for team leads

marlonkeating:
On the About Us page, we have been sorting the members of the DemocracyLab team by role, then by service length.  However, we now want to designate specific team leads, regardless of role title.  To facilitate this, we're introducing a new VolunteerRelation flag 'is_team_lead' that can be set on the django admin side.


PeterBreen (approved):
Checked the appropriate people were flagged as team leads and showed up first, they were and did. Straightforward! 👍

PR #597, opened 2021-04-28
Adding cc as co-owner emails when sending emails to project volunteers

Arpita-22:
Tested with my own email. Please take a look and let me know.

### DIFF
--- a/democracylab/emails.py
+++ b/democracylab/emails.py
@@ -201,6 +201,7 @@ def send_to_project_volunteer(volunteer_relation, subject, template):
         subject=subject,
         from_email=settings.EMAIL_VOLUNTEER_ACCT['from_name'],
         to=[volunteer_relation.volunteer.email],
+        cc=co_owner_emails,
         reply_to=[volunteer_relation.project.project_creator.email] + co_owner_emails,
     )
     email_msg = template.render(email_msg)


marlonkeating (changes requested):
The CC functionality looks good, but there is another step we need to take that I overlooked earlier.

We have a FAKE_EMAILS mode, that when turned on directs all emails to the ADMIN_EMAIL address. 
 In this mode, it's important we don't send anything in the CC field, but do record it for testing purposes. 
 See https://github.com/DemocracyLab/CivicTechExchange/blob/b8ed23785f1f79050344f11764c5a54af233d7a4/democracylab/emails.py#L412 for where we do this.

marlonkeating:
@zh-zhang916 That would work, but for a cleaner approach, you could package the arguments in a dictionary, only add the cc element in the real email case, and then create the email by calling EmailMessage(**arg_dictionary)

Arpita-22:
Added a dummy email string in 'cc', when sending fake emails.

democracylab/emails.py
```diff
@@ -423,6 +424,7 @@ def send_email(email_msg, email_acct=None):
         email_msg.subject = test_email_subject
         email_msg.body = test_email_body
         email_msg.to = [settings.ADMIN_EMAIL]
+        email_msg.cc = ["thisIsADummyEmailAddress",]
```
marlonkeating, line 427: You can just blank this out.  The important thing is to record this information above in test_email_body

marlonkeating (changes requested):

Arpita-22:
updated 'cc' to be blank for fake emails

Arpita-22:
added the 'cc' to  test_email_body.

democracylab/emails.py
```diff
@@ -414,15 +415,17 @@ def send_email(email_msg, email_acct=None):
         email_msg.connection = email_acct['connection'] if email_acct is not None else settings.EMAIL_SUPPORT_ACCT['connection']
     else:
         test_email_subject = 'TEST EMAIL: ' + email_msg.subject
-        test_email_body = '<!--\n Environment:{environment}\nTO: {to_line}\nREPLY-TO: {reply_to}\n -->\n{body}'.format(
+        test_email_body = '<!--\n Environment:{environment}\nTO: {to_line}\nREPLY-TO: {reply_to}\n -->\n{body}\nCC: {cc}'.format(
```
marlonkeating, line 418: cc should come before the {body}

marlonkeating (changes requested):

Arpita-22:
Updated the 'cc' to come before body. Please take a look and let me know.

democracylab/emails.py
```diff
@@ -201,6 +201,7 @@ def send_to_project_volunteer(volunteer_relation, subject, template):
         subject=subject,
         from_email=settings.EMAIL_VOLUNTEER_ACCT['from_name'],
         to=[volunteer_relation.volunteer.email],
+        cc=co_owner_emails,
```
marlonkeating, line 204: We need to include the project owner in this list as well.  You can get the project owner's email with
volunteer_relation.project.project_creator.email

marlonkeating (changes requested):

Arpita-22:
I have added the Project Owner in the 'cc', Please take a look at it and let me know.

PR #599, opened 2021-04-28
Replace facebook tracking with heap tracking

marlonkeating:
A simple replacement of facebook custom event tracking with that of heap.io.


PeterBreen (approved):
Tested, confirmed working.

PR #601, opened 2021-04-29
Update Login form

### LINKED ISSUE #535: Login Page: Sign In button is greyed out if browser auto-populates user name and password
When a browser saves the user name and password and populates it upon navigating to the Login page, the Sign In button is greyed out until some user input is made.

PeterBreen:
- Updates login form to use react-bootstrap form fields like our other forms
 - adds custom validation messages/errors for username and password field
 - fixes #535 by sidestepping the issue; removing the disabled state entirely and validating clientside on submit allows for autofill without issue
 
Still to do if feasible is handling validation on the field level when a field loses focus - e.g. a user types in their email incorrectly and moves on to password, the email field will display the invalid state error immediately, rather than waiting for the form-level validation on submit.

### DIFF
--- a/civictechprojects/static/css/partials/_LogInController.scss
+++ b/civictechprojects/static/css/partials/_LogInController.scss
@@ -19,6 +19,7 @@
 .LogInController-greeting {
   margin-bottom: 27px;
   min-width: 200px;
+  text-transform: uppercase;
 }
 
 .LogInController-signInButton {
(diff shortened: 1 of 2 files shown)


civictechprojects/static/css/partials/_LogInController.scss
```diff
@@ -19,6 +19,7 @@
 .LogInController-greeting {
   margin-bottom: 27px;
   min-width: 200px;
+  text-transform: uppercase;
```
marlonkeating, line 22: What are the advantages of using CSS rather than setting the text as explicitly uppercase?

marlonkeating (commented):
Overall it's a better experience, but I'm noticing the Sign In button fades from green to red, which is an odd visual effect and might wrongly indicate to a user that they are in an error state.

SignInButton

PeterBreen:
> The hover looks good, but the click still goes to orange.

I just took a hammer to that entire mass of CSS and made a `login` button variant, everything is now working off the base green properly.

PR #606, opened 2021-05-06
Companies Page: Learn More urls can be overridden

marlonkeating:
This change updates the companies page Learn More links to read from new environment variable URL_OVERRIDES, which can be used for other urls in the future.


marlonkeating:
After talking with @PeterBreen offline, decided this solution is overkill for what we need.

Issue #469, opened 2020-11-30
Uploading file with long name fails with database error


glenbraun:
Long filenames for uploaded files(over 150 chars) generate s3 keys that exceed the 200 character limit in the database.

We either need to extend the length of the file field, or truncate the key to fit within the 200 character limit.

For an example, create a file with following name and attempt to upload it:
TestFileNameSoLongOver160CharsFileNameSoLongOver160CharsFileNameSoLongOver160CharsFileNameSoLongOver160CharsFileNameSoLongOver160CharsFileNameSoLongOver160.txt

[https://trello.com/c/TP4xbrYR](url)

**Marlon Keating Aug 5, 2019 at 10:35 AM**
Still isn't quite working on the creation page. Saving still fails on the server end.

**Christopher Harrison Jul 22, 2019 at 3:41 PM**
https://github.com/DemocracyLab/CivicTechExchange/pull/252

D-Bits:
@glenbraun The `ProjectFile.file_name` field has a `max_length` of 150 characters. Perhaps would should increase that?

marlonkeating:
Extending the max_length is one option, as is truncating the incoming file name.

PR #608, opened 2021-05-06
Issue #469

D-Bits:
-Double the `max_length` to 300 characters in `ProjectFile.file_name`.


marlonkeating (changes requested):
Missing the migration .py file.  To get that, run
python manage.py makemigrations 
And then apply it locally by running
python manage.py migrate

PR #620, opened 2021-05-12
Revised Docker Volumes

D-Bits:
Added Kent's volume changes, and added CloudFront URL to `example.env` file.


democracylab/settings.py
```diff
@@ -460,10 +450,4 @@ def read_connection_config(config):
 
 PRIVACY_POLICY_URL = os.environ.get('PRIVACY_POLICY_URL')
 
-DONATE_PAGE_BLURB = os.environ.get('DONATE_PAGE_BLURB', '')
-
-GHOST_URL = os.environ.get('GHOST_URL', '')
-GHOST_CONTENT_API_KEY = os.environ.get('GHOST_CONTENT_API_KEY', '')
-
-if GHOST_URL:
-    CSP_CONNECT_SRC = CSP_CONNECT_SRC + (GHOST_URL,)
```
marlonkeating, line 469: We'll need to bring these back in

marlonkeating (changes requested):
Overall this works very well, refreshing backend changes while not erasing the database during the docker-compose up step.  However, it looks like some of the newer settings code from master got lost during the merging process, and that will need to be added back in.

PR #624, opened 2021-05-13
Design System implementation: fonts

### LINKED ISSUE #618: Update Font sizes according to Design System
As part of our design system update and standardization, we want to update how we handle fonts.
https://www.figma.com/file/WADcmVjJh5ARVoZ09xlpfdFN/DemocracyLab?node-id=15075%3A6686

## Web ##
image

## Mobile ##
image

PeterBreen:
Implements design system font stack for mobile and desktop. 

Updates CSS sitewide, consistently applying header classes, adds semantic override classes (headline1, subtitle2) so we can use the proper order of H1->H6 but still have display size flexibility. Several old/unused CSS partials have been removed, and several small related improvements have been made, mostly polish items. One exception to polish items is removing a react key error in the MainHeader component.

For review purposes, we need to look at every page on the site in every state we can think of at every viewport. No big deal, right....?

Closes #618.


marlonkeating (changes requested):
Visual Regressions

## Log In  ## 
image

PeterBreen:
> Visual Regressions
> Log In

This is fixed, but I went a little beyond that with some enhancement: 
On mobile the login button is full width, and the remaining two links below stack in a vertical column. Several of our other pages do that, On tablet and above where there's enough space, they split into the row you'd expect. However, I moved login to the beginning of the row rather than the end so that all the expected actions line up vertically on the left edge - username, then password, then button all aligned, rather than going down for two form fields and fully to the right for the button.

marlonkeating (approved):

PR #627, opened 2021-05-17
Position reorder

### LINKED ISSUE #545: Create Projects: Allow project owners to reorder or hide positions
We want to give project owners the ability to control the order which positions show up on the project cards and project profile page, as well as the ability to hide positions.

# Tasks #
## Backend ##
- [ ] Add database fields for position ordering and visibility
## Create Project Page ##
- [ ] Allow project creator to drag and drop individual positions to reposition them in the list
- [ ] Show a toggle for hiding the position
- [ ] Grey out position when it is hidden
## Project Profile Page ##
- [ ] Only show visible positions
- [ ] Show positions in the order set
## Project Card ##
- [ ] Only show visible positions
- [ ] Show first 3 positions in the order set

image

### LINKED ISSUE #628: Confirmation modals broken if opened a subsequent time
When any of our confirmation modal components are opened a second time, the buttons will be stuck in a 'loading state'.

image

marlonkeating:
Closes #545 

## TODO ##
- [x] Fix #628
- [x] Dim position entry outline when hidden
- [x] Bold position entry when grabbed and moving
- [x] Ask for confirmation before hiding


civictechprojects/static/css/partials/_PositionList.scss
```diff
@@ -0,0 +1,35 @@
...
+  padding-right: 5px;
+}
+i.grip:hover {
+  color:$color-grey-disabled;
+}
+
+.right-side {
+  display: flex;
+  justify-content: space-between;
+  width: 80px;
```
PeterBreen, line 34: Add `margin: auto 0;` to the right-side class and see if that fixes the vertical align issue?

PeterBreen: > 
> 
> Add `margin: auto 0;` to the right-side class and see if that fixes the vertical align issue?

That's a bit too specific, but 
```css
.left-side i, .right-side i {
margin: auto 0;
}
```

should do it.

civictechprojects/static/css/partials/_PositionList.scss
```diff
@@ -0,0 +1,51 @@
...
+  }
+}
+
+.PositionList-entry.dim {
+  color:$color-grey-disabled;
+}
+
+.PositionList-entry.grabbed {
+  cursor: grabbing;
+  background-color: $color-text-grey;
```
PeterBreen, line 26: Rather than the dark text grey for a background, $color-grey-medium may be better here for now - it's more readable when you have the item grabbed. I'm sure we'll be redoing it with a different hover color when the design system gets there but this is good enough for now I think.

civictechprojects/static/css/partials/_PositionList.scss
```diff
@@ -0,0 +1,51 @@
+.PositionList-entry {
+  background-color: $color-background-light;
+  box-shadow: 0 3px 5px rgba(0,0,0,0.2);
+  padding: 5px;
+  margin: 15px;
+  max-width: 600px;
+  display: flex;
+  justify-content: space-between;
```
PeterBreen, line 8: Two things here:
 - Add `align-items: center;` so the role title text ("Business: Analyst") is also vertically centered
 - if the design doc specifies a minimum height, I'd add that too. If not, I'd ...probably still add something like `min-height: 40px;` because these are touch targets on mobile and need to be big enough.

common/components/forms/PositionList.jsx
```diff
@@ -129,40 +211,48 @@ class PositionList extends React.PureComponent<Props, State> {
           message="Do you want to delete this position?"
           onSelection={this.confirmDelete.bind(this)}
         />
+
+        <ConfirmationModal
+          showModal={this.state.showHideModal}
+          message="Do you want to hide this position?"
```
PeterBreen, line 217: Can we get some explanatory text from design on this? Let the user know what hiding means -- it's not deleted but also not shown to the public until you choose to unhide it, whatever. It's not in the specification right now so it's not blocking but I think we should consider it?

common/components/forms/PositionList.jsx
```diff
@@ -129,40 +211,48 @@ class PositionList extends React.PureComponent<Props, State> {
           message="Do you want to delete this position?"
```
PeterBreen, line 211: Is it easy to add the role name to the modal text so it's clearer which role is being deleted? If yes, ask design if they like that idea?

I noticed in testing if I had a list of 5 roles and I clicked delete the modal would come up but which specific delete I clicked on was not clear if I moved my mouse.

PeterBreen (changes requested):
Overall this looks pretty good. The drag and drop works well. I've left some line comments, some of which are suggested improvements and some are bugs that I think need to be fixed. I hope it's clear which is which.

A few other issues that didn't fit in line comments:
- Adding a new role adds it to the bottom of the list, not the top as requested
- Add some space between the grid icon and the role name; it's too close to the icon. 8px looks fine. Red is current, other is proposed.
image
- retitle the page "Which volunteer roles does your project need?" instead of "What type of volunteers..."
- Check if we need conditional rendering as in this comment on figma: 
image

civictechprojects/static/css/partials/_PositionList.scss
marlonkeating: Yep, that color looks much better.

common/components/forms/PositionList.jsx
marlonkeating: Also agree, but it's enough of a lift given the way the code is organized I'd prefer to get design input first.

common/components/forms/PositionList.jsx
marlonkeating: Agree in principle, but think it's something design should chime in on.

PeterBreen: Sounds good.

common/components/forms/PositionList.jsx
PeterBreen: As above, sounds good.

PeterBreen:
Looks good. Initial order and reordering positions both work properly on the cards and profile.

PeterBreen (approved):

PR #640, opened 2021-05-25
Filter out valid project URL path with invalid arguments to canonical url

### LINKED ISSUE #603: Valid project URL path with invalid arguments should show page with canonical url
In our logs we regularly see malformed project urls that are an amalgamation of our old url scheme and our new url scheme.
Example:
/projects/497?section=AboutProject&id=486

When we see a url like this, the request fails with error message
TypeError: must be str, not list

Whether such urls are mistakes or malicious, we should do our best to return a valid page showing the correct canonical url.

zh-zhang916:
I was exploring how our URL system worked, and I think I have got some ideas, credits to @marlonkeating.  We have two versions of the URL system, which the new one contains only the path argument URL but the old one has both. 
The view function for our new URL system is the index function in civictechprojects/views.py. Since the new URL system has only the path argument URL, I filter out all the arguments, started with the '?' mark,  at the beginning of the index function. Tested it with my local machine.

resolve #603

### DIFF
--- a/common/helpers/front_end.py
+++ b/common/helpers/front_end.py
@@ -51,21 +51,27 @@ def get_page_path_parameters(url, page_section_generator=None):
     return match.groupdict()
 
 def clean_invalid_args(url_args):
+    """Filter out invalid query string arguments from old url system
+        Extract args dictionary
+        Remove id and section from dictionary
+        Reconstruct url from dictionary using args_dict_to_string
+
+    Args:
+        url_args(str) : URL query string arguments
+    
+    Returns:
+        str: clean URL query string arguments
+    """
     # Sanity check
     if url_args == "":
         return url_args
     from urllib import parse as urlparse
     # The format of url_args_dict is {'a': ['1'], 'b': ['2']}
     url_args_dict = urlparse.parse_qs(url_args, keep_blank_values=0, strict_parsing=0)
-    new_url_args = ""
-    for key, value in url_args_dict.items():
-        if key == 'section' or key == 'Section':
-            continue
-        if key == 'id' or key == 'Id':
-            continue
-        new_url_args += key + '=' + value[0] + '&'
-
-    return new_url_args[:-1] if new_url_args != "" else new_url_args
+    url_args_dict.pop('section', None)
+    url_args_dict.pop('id', None)
+    url_args_dict = {key : value[0] for key, value in url_args_dict.items()}
+    return args_dict_to_string(url_args_dict)
 
 def get_clean_url(url):
     clean_url = unescape(url)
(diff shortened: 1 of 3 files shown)


zh-zhang916:
> I was exploring how our URL system worked, and I think I have got some ideas, credits to @marlonkeating. We have two versions of the URL system, which the new one contains only the path argument URL but the old one has both.
> The view function for our new URL system is the index function in civictechprojects/views.py. Since the new URL system has only the path argument URL, I filter out all the arguments, started with the '?' mark, at the beginning of the index function. Tested it with my local machine.
> 
> resolve #603

marlonkeating (changes requested):
While this change properly redirects invalid links like /projects/497?section=AboutProject&id=486, it also redirects valid links like:
/projects/1?embedded=1
/projects?sortField=-project_date_modified&issues=education

I would put the logic for this in a separate method, and write unit tests to easily test that it satisfies all the known scenarios (see https://github.com/DemocracyLab/CivicTechExchange/blob/4b8c5038422faf2fa51ef008ea67ea9d848ed060/common/tests/test_helpers.py for testing examples)

common/helpers/front_end.py
```diff
@@ -50,6 +50,22 @@ def get_page_path_parameters(url, page_section_generator=None):
     match = page_section_generator['regex'].search(url)
     return match.groupdict()
 
+def clean_invalid_args(url_args):
```
marlonkeating, line 53: A cleaner way to handle the logic in this method would be to
- Extract args dictionary
- Remove id/section from dictionary
- Reconstruct url from dictionary using args_dict_to_string

marlonkeating, line 53: Be sure to document the behavior of this method and what it returns

marlonkeating (changes requested):

common/helpers/front_end.py
zh-zhang916: I am not sure if it is the right way to document it, but I put down behaviour, args and return using standard docstrings.

zh-zhang916: Thanks for pointing out that function for me to use. I have incorporated it and write the code in a much cleaner way.

zh-zhang916:
Tested it with my local machine. Work as expected.

common/tests/test_helpers.py
```diff
@@ -46,6 +46,12 @@ def test_get_clean_url(self):
         expected = '&clean'
         self.assertEqual(expected, get_clean_url('&amp;clean'))
 
+    def test_clean_invalid_args(self):
```
marlonkeating, line 49: Should add some test cases for urls with section and id too.

marlonkeating (changes requested):

marlonkeating (approved):

PR #644, opened 2021-06-02
Beth/sponsor footer

bhg:


marlonkeating (changes requested):
You'll need to include the migration file for this, which you generate with:
python manage.py makemigrations

bhg:
> You'll need to include the migration file for this, which you generate with:
> python manage.py makemigrations

Added this now.

common/components/componentsBySection/CreateEvent/AboutEventDisplay.jsx
```diff
@@ -118,6 +119,7 @@ class AboutEventDisplay extends React.PureComponent<Props, State> {
         {!_.isEmpty(event.event_legacy_organization) && (
           <ProfileProjectSearch viewOnly={this.props.viewOnly} />
         )}
+        <MainFooter key="main_footer" forceShow={event.show_headers}/>
```
marlonkeating, line 122: The MainFooter component was designed to be full-width, and inside the "AboutEvent-root container" div puts it within a smaller frame.  I would either put it in the AboutEventController component, or just outside the div in this component.

marlonkeating (changes requested):

common/components/componentsBySection/CreateEvent/AboutEventDisplay.jsx
bhg: Done and tested.

common/components/chrome/MainFooter.jsx
```diff
@@ -18,9 +18,13 @@ const sectionsToShowFooter: $ReadOnlyArray<string> = [
...
 ];
 
-class MainFooter extends React.Component<{||}> {
-  constructor(): void {
-    super();
+type Props = {|
+  forceShow: boolean,
+|};
+
+class MainFooter extends React.Component<{||}, Props> {
```
marlonkeating, line 25:
```suggestion
class MainFooter extends React.Component<Props> {
```

marlonkeating (approved):

PR #650, opened 2021-06-09
Allanbranch2

allan343:
Hi Marlon,

   I don't see where you can add a title and description for pull requests.

This is for users showing up as active in my projects before they were approved.  Now they show up as pending if they are not approved.

I tried to tidy up my formatting by right clicking and doing "format document"

### DIFF
--- a/common/components/componentsBySection/MyProjects/MyProjectCard.jsx
+++ b/common/components/componentsBySection/MyProjects/MyProjectCard.jsx
@@ -61,49 +61,39 @@ class MyProjectCard extends React.PureComponent<Props, State> {
   }
 
   _renderProjectStatus(): React$Node {
-    /*
-    const header: string = this.state.isOwner
-      ? "Project Status"
-      : "Volunteer Status";
-    const status: string = this.state.isOwner
-      ? this.props.project.isApproved
-        ? "Published"
-        : "Unpublished"
-      : !this.props.project.isApproved
-      ? "Awaiting Approval"
-      : this.props.project.isUpForRenewal
-      ? "Expires on " + moment(this.props.project.projectedEndDate).format("l")
-      : "Active";
-      */
-     let header = ""
-    if(this.state.isOwner){
-      header ="Project Status";
+
+    let header = ""
+
+    if (this.state.isOwner) {
+      header = "Project Status";
     }
-    else{
+
+    else {
       header = "Volunteer Status";
     }
-      let status ="";
 
-    if(this.state.isOwner)
-    {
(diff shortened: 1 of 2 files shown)


common/components/componentsBySection/MyProjects/MyProjectCard.jsx
```diff
@@ -61,18 +61,40 @@ class MyProjectCard extends React.PureComponent<Props, State> {
...
+
+    let header = ""
+
+    if (this.state.isOwner) {
+      header = "Project Status";
+    }
+
+    else {
+      header = "Volunteer Status";
+    }
```
marlonkeating, line 73:
```suggestion
const header: string = this.state.isOwner ? "Project Status"  :  "Volunteer Status";
```
Using a conditional operator for a simple if/else value assignment is fine, and in many cases like this I think enhances readability.  It's just when we get to multi-pronged nested conditionals that readability becomes a real issue for this construction.

common/components/componentsBySection/MyProjects/MyProjectCard.jsx
```diff
@@ -61,18 +61,40 @@ class MyProjectCard extends React.PureComponent<Props, State> {
...
+      status = "Expires on " + moment(this.props.project.projectedEndDate).format("l");
+    }
+
+    else if (!this.state.isOwner && !this.props.project.isApproved) {
+      status = "Pending";
+    }
+
+    else {
+      status = "Active";
+    }
```
marlonkeating, line 96:
```suggestion
    if (this.state.isOwner) {
      if (this.props.project.isApproved) {
        status = "Published";
      } else {
        status = "Unpublished";
      }
    } else if (this.props.project.isUpForRenewal) {
      status = "Expires on " + moment(this.props.project.projectedEndDate).format("l");
    } else if (!this.state.isOwner && !this.props.project.isApproved) {
      status = "Pending";
    } else {
      status = "Active";
    }
```
It's easier to see that these code blocks are part of a single conditional check if they're attached to each other rather than on separate lines.

marlonkeating (changes requested):
For tidying up your formatting, it doesn't look like you have the prettier formatter hooked up to VS Code's Format Document command.  If you want to do that, make sure you have the [Prettier Extension](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode) installed, and then set your default formatter to the prettier extension.
image

PeterBreen:
Here's how to change a PR title: 
image

there may be other ways but that's where I do it :)

allan343:
closing to open a new pr

PR #665, opened 2021-06-17
Increased ProjectFile.file_name max_length

D-Bits:
Migration file included.


civictechprojects/models.py
```diff
@@ -780,7 +780,7 @@ class ProjectFile(models.Model):
     file_group = models.ForeignKey(Group, related_name='files', blank=True, null=True, on_delete=models.CASCADE)
     file_event = models.ForeignKey(Event, related_name='files', blank=True, null=True, on_delete=models.CASCADE)
     file_visibility = models.CharField(max_length=50)
-    file_name = models.CharField(max_length=150)
+    file_name = models.CharField(max_length=300)
     file_key = models.CharField(max_length=200)
```
marlonkeating, line 784: file_key is based on file_name, so it should be at least 300 chars too

PR #666, opened 2021-06-17
User Profile: Reorganize into 2-column layout

marlonkeating:
This is the first stage of the User Profile revamp, reformatting the page into a 2-column format.


PeterBreen (changes requested):
I think I've caught all the big front-end style changes needed. I'm at the point where I think we need to make them and see what needs polishing up in a second pass, if anything. It's getting hard to track in devtools :)

Let me know if any of the line comments are unclear and/or you want more detail; I wasn't sure how much to put in.

common/components/controllers/AboutUserController.jsx
```diff
@@ -51,78 +53,97 @@ class AboutUserController extends React.PureComponent<{||}, State> {
...
-        <div className="AboutProjectController-root">
-          <div className="container">
-            <div className="row background-light about-user-section">
-              <div className="col-12">
-                <Avatar user={user} size={50} />
-                <h1>{user && user.first_name + " " + user.last_name}</h1>
-              </div>
+        <div className="AboutUser-root">
+          <div className="row background-light about-user-section">
+            <div className="col-12 col-lg-2 left-column">
```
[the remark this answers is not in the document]
marlonkeating: Logged as https://github.com/DemocracyLab/CivicTechExchange/issues/674

civictechprojects/static/css/partials/_AboutUser.scss
```diff
@@ -1,16 +1,55 @@
+.AboutUser-root .left-column {
+  border-right: solid;
+  border-width: 1px;
+  border-color: $color-disabled-border;
```
PeterBreen, line 4: This should only apply in a media query for desktop; the border shows up on mobile/tablet where it's unintended:

image

(hard to see but that border right which stops partway down is from this rule)

civictechprojects/static/css/partials/_AboutUser.scss
```diff
@@ -11,6 +56,8 @@
     text-decoration: underline;
   }
```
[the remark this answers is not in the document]
PeterBreen: Did removing this rule break something in the page? It should just fall back to site default, is it not?

civictechprojects/static/css/partials/_IconLinkDisplay.scss
```diff
@@ -25,14 +25,18 @@
...
   line-height: 23px;
   border-top: 1px solid $color-grey-medium;
   border-top-right-radius: 5px;
   border-right: 1px solid $color-grey-medium;
   border-bottom-right-radius: 5px;
   border-bottom: 1px solid $color-grey-medium;
+  display: block;
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
```
PeterBreen, line 39: I think this works for now, let's call it good. Further fixes to IconLink seem best in another PR because they show up on multiple pages so I don't want them to be blocking. If you want to roll up a full iconlink fix in this PR though let me know, and I can push a commit or something.

common/components/controllers/AboutUserController.jsx
```diff
@@ -51,78 +53,97 @@ class AboutUserController extends React.PureComponent<{||}, State> {
...
-  _renderTechnologies(): ?Array<React$Node> {
-    const user: UserAPIData = this.state.user;
+  _renderAboutMe(user: UserAPIData): React$Node {
     return (
-      user &&
-      user.user_technologies && (
-        <TagsDisplay tags={user && user.user_technologies} />
-      )
+      <div className="about-user-section">
+        <div className="col-12">
```
PeterBreen, line 118: since you removed the `row` class here, should also remove the `col` (they need each other to work properly, and neither is needed here)

common/components/controllers/AboutUserController.jsx
```diff
@@ -51,78 +53,97 @@ class AboutUserController extends React.PureComponent<{||}, State> {
...
+      </div>
     );
   }
 
-  _renderLinks(): ?Array<React$Node> {
-    const user: UserAPIData = this.state.user;
+  _renderAreasOfInterest(user: UserAPIData): React$Node {
+    return (
+      <div className="about-user-section">
+        <div className="col-12">
```
PeterBreen, line 132: as above, can remove this col-12

common/components/controllers/AboutUserController.jsx
```diff
@@ -51,78 +53,97 @@ class AboutUserController extends React.PureComponent<{||}, State> {
...
+        {showEdit && this._renderEditUserButton()}
+        <div className="about-user-section">
+          <Avatar user={user} imgClass="Profile-img" />
+        </div>
+        <div className="about-user-section">
+          <h3>{user && user.first_name + " " + user.last_name}</h3>
+        </div>
+        {!_.isEmpty(user.user_links) ? (
+          <div className="about-user-section">
+            <div className="col-12">
```
PeterBreen, line 84: can remove this col-12 (missed it at first, explanation below)

common/components/controllers/AboutUserController.jsx
```diff
@@ -51,78 +53,97 @@ class AboutUserController extends React.PureComponent<{||}, State> {
...
-                </div>
-              </div>
-            ) : null}
+        {user && !_.isEmpty(user.user_technologies)
+          ? this._renderAreasOfInterest(user)
+          : null}
+
+        {user && !_.isEmpty(user.user_files) ? (
+          <div className="about-user-section">
+            <div className="col">
```
PeterBreen, line 105: can remove this "col" (which is a valid Bootstrap class, but not needed here)

PeterBreen (changes requested):
Looking good, few more items and then I think we're set. 🎉 

One item I didn't put in a line comment for:
image

The content is right up on the footer, and on mobile the profile image is right up against the nav. There's 15px on the project profile to give just a bit of space and it works well.

quickest fix is probably this:
```css
.AboutUser-root .left-column {
  padding-top: 15px;
}
.AboutUser-root .right-column {
  padding-bottom: 15px;
}
```

civictechprojects/static/css/partials/_IconLinkDisplay.scss
marlonkeating: This currently works on all the existing pages.  We can do more comprehensive fixes in a separate PR.

PeterBreen (approved):

PR #673, opened 2021-06-22
Removed prerender code.

D-Bits:
Included JSX prerender code.


democracylab/settings.py
```diff
@@ -100,8 +99,8 @@
 
 # TODO: Use     'django_seo_js.middleware.UserAgentMiddleware',
 MIDDLEWARE = [
-    'common.helpers.caching.DebugUserAgentMiddleware',
     'django.middleware.security.SecurityMiddleware',
+    'django_seo_js.middleware.UserAgentMiddleware',
```
marlonkeating, line 103: Should remove this and the dependency in https://github.com/DemocracyLab/CivicTechExchange/blob/4b420b49872edf9fc3a87828f829ccbba65eece0/requirements.txt#L32

democracylab/settings.py
```diff
@@ -215,7 +214,7 @@ def read_connection_config(config):
 EMAIL_BACKEND = 'django.core.mail.backends.console.EmailBackend'
 
 PROTOCOL_DOMAIN = os.environ['PROTOCOL_DOMAIN']
-ADMIN_EMAIL = os.environ['ADMIN_EMAIL']
+ADMIN_EMAIL = os.getenv('ADMIN_EMAIL')
```
marlonkeating, line 217: Any reason for this change?  ADMIN_EMAIL is a required environment variable, so it's okay if a null variable throws an exception

democracylab/settings.py
```diff
@@ -400,9 +399,6 @@ def read_connection_config(config):
 
 # If you're using prerender.io (the default backend):
 SEO_JS_PRERENDER_TOKEN = os.environ.get('SEO_JS_PRERENDER_TOKEN', '')
-SEO_JS_BACKEND = "common.helpers.caching.DebugPrerenderIO"
```
marlonkeating, line 403: Anything prefixed with SEO_JS_ is something we should be able to get rid of.

marlonkeating (changes requested):

democracylab/settings.py
D-Bits: I got a `ValueError: Missing SEO_JS_PRERENDER_TOKEN in settings.` error when I tried to boot the containers after removing those values.

marlonkeating: Did you try it after removing the library references from requirements.txt and MIDDLEWARE?

PR #676, opened 2021-06-24
IconLink component layout fixes

PeterBreen:
- Updates IconLink layout and sizing rules so it'll conform to parent size instead of fixed width (which broke a lot on responsive viewport changes)
- Long titles are truncated instead of wrapping/breaking the element
- Link colors are represented instead of black text with no hover change, so the link part of IconLink look like links

Before:
image
Fixed width, regardless of available space from parent element

After:
image

Scales to fit available space, link colors are consistent with the rest of the site, and fixes a misaligned H4, bonus.

On tablets these links may look a little weird due to stretching full width but why not use the space if it's available, right?

### DIFF
--- a/civictechprojects/static/css/partials/_IconLinkDisplay.scss
+++ b/civictechprojects/static/css/partials/_IconLinkDisplay.scss
@@ -8,8 +8,8 @@
   flex-basis: 100%;
   flex-wrap: nowrap;
   min-width: 0; // see https://css-tricks.com/flexbox-truncated-text/
+  text-decoration: none; // we decorate sub-elements to control what does it
 }
-
 
 .IconLink-root p {
   margin: 0;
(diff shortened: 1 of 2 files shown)


marlonkeating (changes requested):
The autofit behavior works great, the only issue  I can see is that the glyphs are getting underlined along with the link text when hovered over.

image

civictechprojects/static/css/partials/_IconLinkDisplay.scss
```diff
@@ -46,7 +46,7 @@
     text-overflow: ellipsis;
   }
 }
-.IconLink-right:hover, .IconLink-right:active {
+.IconLink-root:hover p, .IconLink-root:active p {
```
PeterBreen, line 49: Making it so the hover state applies no matter where your mouse cursor is, previously it wouldn't work properly if you hovered over the icon on the left rather than the text on the right. Gives better hinting that the whole box is clickable this way.

marlonkeating (approved):

PR #677, opened 2021-06-25
When a user is not logged in, looking at available postions a "log in to apply" button will appear

allan343:
Hi,  

   For this pull request, I had to move showApplyButton as a pass in variable to _renderApplyButton().  The reason for this was, the way it was written before:

  {showApplyButton && this._renderApplyButton()}

showApplyButton kept evaluating to not true when user is not logged in.  This being the case _renderApplyButton() was never run.  When _renderApplyButton() does not run my new condition:

 if (!CurrentUser.isLoggedIn())

 is never evaluated and my new button is never rendered.

  The first time I had this working I simply commented out showApplyButton resulting in  {/*showApplyButton && */this._renderApplyButton()}. 

Can you take a look at what I did?


common/components/common/positions/AboutPositionEntry.jsx
```diff
@@ -55,9 +55,9 @@ class AboutPositionEntry extends React.PureComponent<Props> {
     );
   }
 
-  _renderApplyButton(): ?React$Node {
+  _renderApplyButton(showApplyButton): ?React$Node {
```
marlonkeating, line 58:
```suggestion
  _renderApplyButton(showApplyButton: boolean): ?React$Node {
```

common/components/common/positions/AboutPositionEntry.jsx
```diff
@@ -68,15 +68,18 @@ class AboutPositionEntry extends React.PureComponent<Props> {
           Apply Now
         </Button>
       );
-    } else if (!CurrentUser.isLoggedIn()) {
+    }
+    if (!CurrentUser.isLoggedIn()) {
```
marlonkeating, line 72: Any reason you changed this from an 'else if' construction?

marlonkeating (changes requested):

marlonkeating (changes requested):

marlonkeating (approved):

PR #678, opened 2021-07-03
Fix front-end redirection after login

### LINKED ISSUE #671: Login: Enable redirection back to pages with embedded id's
When the user goes to log in, they are redirected back to the page they were on.  However, this doesn't work for pages which have an embedded id.

Where login url is set on the login button:
https://github.com/DemocracyLab/CivicTechExchange/blob/4b420b49872edf9fc3a87828f829ccbba65eece0/common/components/chrome/MainHeader.jsx#L57

Login controller:
https://github.com/DemocracyLab/CivicTechExchange/blob/4b420b49872edf9fc3a87828f829ccbba65eece0/common/components/controllers/LogInController.jsx#L26

Where redirection happens on back-end:
https://github.com/DemocracyLab/CivicTechExchange/blob/4b420b49872edf9fc3a87828f829ccbba65eece0/democracylab/views.py#L26

marlonkeating:
closes #671

### DIFF
--- a/common/components/chrome/MainHeader.jsx
+++ b/common/components/chrome/MainHeader.jsx
@@ -209,7 +209,6 @@ class MainHeader extends React.Component<{||}, State> {
   }
 
   //TODO: Refactor these to reduce duplication
-  //TODO: Allow multiple arguments for url.section to handle url.getPreviousPageArg() - options object?
   _renderNavLink(section, text, classes = "") {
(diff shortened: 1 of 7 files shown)


PeterBreen (approved):
Tested this, looks good. We know it doesn't fix all the redirection issues but the cases it tackles, it works great 👍

PR #679, opened 2021-07-05
URL encode names of new projects and events when passing in URL, and …

bhg:
…decode in alert header.

### DIFF
--- a/common/components/chrome/AlertHeader.jsx
+++ b/common/components/chrome/AlertHeader.jsx
@@ -37,14 +37,12 @@ type State = {|
 
 const AlertMessages: Dictionary<string> = {
   projectAwaitingApproval:
-    'Your project ',
+    'Your project "${value}" is awaiting approval.  Expect a decision in the next business day.',
   eventAwaitingApproval:
-    'Your event ',
+    'Your event "${value}" is awaiting approval.  Expect a decision in the next business day.',
   groupAwaitingApproval:
-    'Your group ',
+    'Your group "${value}" is awaiting approval.  Expect a decision in the next business day.',
 };
-
-const AlertMessageEnding: string = ' is awaiting approval.  Expect a decision in the next business day.';
 
 class AlertHeader extends React.Component<Props, State> {
   constructor(): void {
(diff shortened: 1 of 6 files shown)


bhg:
I decided in the end that passing project or event ID and fetching the title would involve an enormous amount of effort and obfuscation, and that simply encoding the URL correctly would be simpler, more intuitive and run faster.

I also noticed here that adding a new Group does not bring up an alert banner as do Projects and Events. Is this by design or should this be added?

marlonkeating (changes requested):
This is a big improvement, but I saw at least one test case it didn't handle quite right

image

Repro query: projectAwaitingApproval=Test%20Project%20%24%26%2B%23%20/%5C-

I would recommend putting the encode/decode steps into helper methods that can be unit tested.

We do need to show this for the Create Group scenario, and I added an issue for that: https://github.com/DemocracyLab/CivicTechExchange/issues/680
If you'd like to wrap that fix into this PR you can.

bhg:
> This is a big improvement, but I saw at least one test case it didn't handle quite right
> 
> image
> 
> Repro query: projectAwaitingApproval=Test%20Project%20%24%26%2B%23%20/%5C-
> 
Looks like this is happening because the replace() function is interpreting parts of the project name as a regex. There's actually no need to use replace() here - a simple concatenation would get rid of this weird behavior.

bhg:
Implemented the above suggestions. The only thing missing right now is running the new tests to ensure that they passed.

common/components/chrome/AlertHeader.jsx
```diff
@@ -178,10 +182,8 @@ class AlertHeader extends React.Component<Props, State> {
 
   _renderTriggeredAlert(): React$Node {
     const key: string = _.keys(AlertMessages).find(key => url.argument(key));
-    const message: string = AlertMessages[key].replace(
-      "{value}",
-      decodeURI(url.argument(key))
-    );
+    const message: string = AlertMessages[key] + '"' + url.decodeNameFromUrlPassing(url.argument(key)) + '"' 
+                          + AlertMessageEnding;
```
marlonkeating, line 186: Avoiding duplicating code is generally good, but in this case I think the added code complexity to reduce duplication, coupled with difficulty future code editors might have in searching for these now chopped up front-end strings, would make me prefer the old approach.

bhg: Agreed, and I did this to deal with Javascript's strange replace() functionality, which interprets sequences such as $& as flags.

In the end I realized that I can very easily void this by just performing the string replacement when the project/event/group name is still URL encoded.

marlonkeating (approved):

PR #681, opened 2021-07-06
Remove leftover prerender settings.

D-Bits:
From `settings.py`.


marlonkeating (changes requested):
Almost there, just one thing to remove.

https://github.com/DemocracyLab/CivicTechExchange/blob/0ae97fef7de82029e05f5c7312552c4ec916ee2d/civictechprojects/templates/new_index.html#L29-L32

It would also be good to rename these tests so that they focus on the is_sitemap_url check instead of prerendering.

https://github.com/DemocracyLab/CivicTechExchange/blob/0ae97fef7de82029e05f5c7312552c4ec916ee2d/common/tests/test_helpers.py#L11-L23

PR #688, opened 2021-07-10
Donate Page: Change default donation to monthly

marlonkeating:
## Changes ##
- Monthly $25 is now the default donation option
- Added $5 option and removed $75 option


PeterBreen (approved):
Looks good, correctly displays the default selected values. 👍

PR #689, opened 2021-07-12
Allan array test

allan343:
Hi,

   My unit test for array.js

Thanks,

Allan


marlonkeating (changes requested):
I appreciate the care that went into the code, in terms of typing out everything with flow and using objects that are similar to what we use in production.

However, for test code it's best to use the minimal amount of code that will test the functionality in question.  It is entirely possible to test the array.join function in a single line of code (which I don't necessarily recommend; personally I would write the test in the fewest lines of code that feels readable to me).

marlonkeating:
Done as part of https://github.com/DemocracyLab/CivicTechExchange/pull/690

PR #690, opened 2021-07-14
Unit test for array

allan343:
I'm having a hard time with this pr.  It keeps showing me changes I did not make.  Can one of you give me some help?


marlonkeating (approved):
The unexpected changes you see are a consequence of running prettier on code that did not have prettier run on it before, and is totally fine.  The test runs fine, and is much cleaner than before.

PR #691, opened 2021-07-14
Unit tests for utils.js.

### LINKED ISSUE #633: Add Unit Tests: utils.js
We need unit tests for https://github.com/DemocracyLab/CivicTechExchange/blob/master/common/components/utils/utils.js

zh-zhang916:
As the title said.

close #633

### DIFF
--- a/common/components/test/utils.test.js
+++ b/common/components/test/utils.test.js
@@ -7,6 +7,7 @@ import window from "./__mocks__/window";
 import NavigationStore from "../stores/NavigationStore.js";
 import renderer from "react-test-renderer";
 import GroupBy from "../utils/groupBy.js";
+import utils from "../utils/utils.js";
 
 describe("utils", () => {
   test("async.doWhenReady", () => {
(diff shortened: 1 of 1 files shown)


common/components/test/utils.test.js
```diff
@@ -167,5 +168,30 @@ describe("utils", () => {
...
+    expect(tranformedStringHtmlContent).toEqual("<p>Function unescapeHtml's test on <a href='fake url'>something</a>something</p>");
+  }); 
+
+  test("utils.pluralize", () => {
+    let word = utils.pluralize("apple", "apples", 1);
+    expect(word).toEqual("apple");
+    word = utils.pluralize("apple", "apples", '1');
+    expect(word).toEqual("apples");
+    word = utils.pluralize("apple", "apples", [1]);
+    expect(word).toEqual("apples");
```
marlonkeating, line 191: While it can be good to test functions with unexpected input, with unit tests we should stick with defined behavior.  The 'count' variable is expected to be a number, and if other types are used we don't have defined behavior for them.  Testing against strings or arrays are essentially testing our implementation details, not the contract of the function itself, and so I would remove these tests.

marlonkeating (changes requested):

common/components/test/utils.test.js
zh-zhang916: Thank you for your suggestions. Removed :).

marlonkeating (approved):

PR #693, opened 2021-07-15
Video page: Add embedded youtube video

marlonkeating:
This change creates a new video page for showing to prospects, with the url controlled by environment variable YOUTUBE_VIDEO_URL

### DIFF
--- a/common/components/controllers/VideoController.jsx
+++ b/common/components/controllers/VideoController.jsx
@@ -10,15 +10,17 @@ class VideoController extends React.PureComponent {
 
   render(): React$Node {
     return (
-      <iframe
-        width="560"
-        height="315"
-        src="https://www.youtube.com/embed/nvIUWtx-nmo"
-        title="YouTube video player"
-        frameBorder="0"
-        allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture"
-        allowFullScreen
-      ></iframe>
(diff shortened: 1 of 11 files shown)


common/components/controllers/VideoController.jsx
```diff
@@ -0,0 +1,26 @@
...
+
+// Shows embedded video
+class VideoController extends React.PureComponent {
+  constructor(): void {
+    super();
+  }
+
+  render(): React$Node {
+    return (
+      <iframe
```
PeterBreen, line 13: I'd suggest wrapping this iframe in a <div> so we can do some positioning. Assuming the div is named something like youtube-video-container, then:

```css
.youtube-video-container {
  position: relative;
  overflow: hidden;
  width: 95%;
}

.youtube-video-container::after {
  display: block;
  content: "";
  padding-top: 56.25%;
}

.youtube-video-container iframe {
  position: absolute;
  top: 0;
  left: 0;
  width: 100%;
  height: 100%;
}
```

I just searched 'responsive youtube embed' or something similar and came up with [this article](https://techstacker.com/how-to-responsive-youtube-videos/) but I'm sure there are others. My goal was to center it, enlarge it, and then push it off the footer a bit because it's really big.

It's not perfect but I think it's an improvement. Try it out and see if you like it?

PeterBreen (approved):
Works at every viewport I tested, no issues.

PR #695, opened 2021-07-19
Shorten long file names with ...

HebuPatil:
Shortens long file names that are larger than 35% of card width and ends it with "...". Only changes display of file name not file name itself.

### DIFF
--- a/civictechprojects/static/css/partials/_FileUploadList.scss
+++ b/civictechprojects/static/css/partials/_FileUploadList.scss
@@ -1,0 +1,15 @@
+.FileUploadList-item {
+    display: flex;
+    align-items: center;
+    i {
+        margin-left: 1rem;
+    }
+    a {
+        max-width: calc(100% - 40px);
+        display: inline-block;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+    }
+
+}
--- a/civictechprojects/static/css/styles.scss
+++ b/civictechprojects/static/css/styles.scss
@@ -97,3 +97,4 @@
 @import "partials/PositionList";
 @import "partials/LinkList";
 @import "partials/Video";
+@import "partials/FileUploadList";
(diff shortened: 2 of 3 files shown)


PeterBreen (changes requested):
Hey, nice work so far! This is really close, and your truncation CSS classes are spot on, but I think maxWidth: 35% is too constrained and DLab does our CSS differently. 

That maxwidth rule leaves a lot of empty usable space that we don't need to truncate on large devices, and on smaller devices with limited space available it leads to over-aggressive truncation, like so:
image

What I'd suggest is setting a max width which will let the length of the file name determine whether or not it truncates; so rather than max-width: 35%, try  `max-width: calc(100% - 40px);`  (the -40px in this calc is so the trash can icon has space no matter what)

Also, DemocracyLab doesn't use CSS-in-JS. All our styles are contained in SASS partials in `/civictechprojects/static/css/partials/` so if you could replace the inline style with classnames, create a partial, and put the CSS there that'd be appreciated. To do that,
 - create `_FileUploadList.scss` in /civictechprojects/static/css/partials/ and add appropriate classes, CSS rules
  - add `@import "partials/FileUploadList";` to /civictechprojects/static/css/styles.scss - at the end of the file is fine for this
  - update the JSX to use those classes
  
  Our class naming scheme is `ComponentName-descriptor` so in this case something like `FileUploadList-link` or `FileUploadList-item` would be fine, though feel free to come up with better names if something occurs to you.
  

### Optional Extra Change

If you're willing, we have some pre-existing display issues in this file that would be a fairly quick fix since you're in it already. Please note you do not have to take this on, but if you do: Take a look at the trash can icon and see it's both too close to the text and not vertically aligned properly. 


To fix the align, I'd suggest flexbox on the div in L138: 
```
display: flex; 
align-items: center;
```
And then to space it out better, add this to the icon on L142:
```
margin-left: 1rem; 
```
Those two combined ought to get it looking even better. 

Please let me know if you have any questions about any part of this. Thanks!

PeterBreen (approved):

PeterBreen:
:100:

Looks great now!

PR #701, opened 2021-07-21
Unit test for guard.js.

### LINKED ISSUE #631: Add Unit Tests: guard.js
We need unit tests for https://github.com/DemocracyLab/CivicTechExchange/blob/master/common/components/utils/guard.js

zh-zhang916:
As the title says.

close #631

### DIFF
--- a/common/components/test/utils.test.js
+++ b/common/components/test/utils.test.js
@@ -200,9 +200,14 @@ describe("utils", () => {
   }); 
 
   test("guard.duplicateInput", () => {
-    _.debounce = jest.fn();
-    const testFunc = () => "test";
-    Guard.duplicateInput(testFunc);
-    expect(_.debounce).toBeCalledWith(testFunc, 1000, {leading: true});
+    const testFunc = jest.fn();
+    expect(testFunc).toHaveBeenCalledTimes(0);
+    const func = Guard.duplicateInput(testFunc);
+    func();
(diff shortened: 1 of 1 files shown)


common/components/test/utils.test.js
```diff
@@ -196,4 +198,11 @@ describe("utils", () => {
     word = utils.pluralize("apple", "apples", 5);
     expect(word).toEqual("apples");
   }); 
+
+  test("guard.duplicateInput", () => {
+    _.debounce = jest.fn();
+    const testFunc = () => "test";
+    Guard.duplicateInput(testFunc);
+    expect(_.debounce).toBeCalledWith(testFunc, 1000, {leading: true});
```
marlonkeating, line 206: I would instead test that testFunc was called only once after multiple invocations of Guard.duplicateInput's return function.

marlonkeating (changes requested):
The current test is only testing the implementation details, where we want to test the behavior regardless if the implementation changes in the future.

marlonkeating (changes requested):
Test looks good so far.  There's a couple more things we should test

## Checklist ##
- [x] Verify return value of function is duplicated
- [ ] Verify function is called again after timeout period elapses

zh-zhang916:
Fixed.

marlonkeating (approved):

PR #702, opened 2021-07-22
Remove prerender

D-Bits:
Hopefully, this will be the last of the changes.


common/tests/test_helpers.py
```diff
@@ -9,12 +9,12 @@
 
 
 class CommonHelperTests(TestCase):
-    def test_prerender_urls(self):
+    def test_is_sitemap_urls(self):
         urls = SitemapPages()
         for url in urls:
             self.assertTrue(is_sitemap_url(url), 'Should be able to prerender ' + url)
```
marlonkeating, line 15: Should update test messages too.

marlonkeating (changes requested):
Can also remove references from

https://github.com/DemocracyLab/CivicTechExchange/blob/ec23e41905d33f92fe07489f50fed2b2717bc2af/democracylab_environment_variables.sh#L114-L115

common/tests/test_helpers.py
D-Bits: Just finished removing those from both files.

common/components/controllers/AboutProjectController.jsx
```diff
@@ -90,7 +88,6 @@ class AboutProjectController extends React.PureComponent<{||}, State> {
         <React.Fragment>
           <Helmet>
             <meta
-              name="prerender-status-code"
```
marlonkeating, line 93: This whole section should be removed.

marlonkeating (changes requested):

common/components/controllers/AboutProjectController.jsx
D-Bits: Just removed it.

marlonkeating (approved):

Issue #540, opened 2021-02-06
My Projects: Buttons are cut off on 320px mobile view


marlonkeating:
The action buttons on owned projects are cut off on a 320 px wide mobile view.

![image](https://user-images.githubusercontent.com/322346/107130463-8b7a0900-6882-11eb-8883-bc742a8e1f98.png)

PR #706, opened 2021-07-27
My projects button cutoff at 320px mobile view

allan343:
This is for issue https://github.com/DemocracyLab/CivicTechExchange/issues/540

This is the solution I came up with but there are some definite tradeoffs.  Tell me what you think.

### DIFF
--- a/civictechprojects/static/css/partials/_MyProjectCard.scss
+++ b/civictechprojects/static/css/partials/_MyProjectCard.scss
@@ -27,8 +27,8 @@
 }
 
 @media screen and (max-width: 370px) {
-  .MyProjectCard-column-ProjectName {
-    max-width: 40px;
+  .MyProjectCard-column-ProjectName,.MyProjectCard-column-ProjectHeader,.MyProjectCard-column-ProjectStatus {
+    max-width: 70px;
     overflow: hidden;
     text-overflow: ellipsis;
     white-space: nowrap;
--- a/civictechprojects/static/css/partials/_MyProjectCard.scss
+++ b/civictechprojects/static/css/partials/_MyProjectCard.scss
@@ -26,13 +26,11 @@
   margin: 5px;
 }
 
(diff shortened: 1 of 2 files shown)


allan343:
If you don't like what I'm doing.  I was hoping @marlonkeating  or @PeterBreen  can give me a suggestion as to what to do.  Basically I noticed the buttons were clipped at 320px because the contents of the columns to the left of the buttons were pushing it toward edge causing the clipping.  My change involves making those columns smaller.

PeterBreen (changes requested):
Hi Allan, I took a look at this and I think your solution is workable - but I agree with you that there may be another way. What I'd suggest is making the table cells stack vertically on mobile (using inline-block), then reverting to standard row behavior on larger screens. That should let us skip truncation entirely. 

Also, my preference for dlab's style is to avoid max-width media queries if possible -- they're perfectly valid, but our pattern is to set mobile rules as the base and then override as needed with min-width rules as the viewport expands. 

 So I'd refactor this to be something like
 ```
.MyProjectCard-table td {
  display: inline-block;
  width: 100%;
}
```
and then to go back to standard behavior on larger devices:
```
@include media-breakpoint-up(md) {
  .MyProjectCard-table td {
    display: table-cell;
    width: 250px;
    }
}
```


If you're not familiar, media-breakpoint-up is [a sass mixin](https://getbootstrap.com/docs/4.6/layout/overview/#responsie-breakpoints), it's just another way to write a media query that also keeps our responsive breakpoints and style changes synced up. I'm using the medium breakpoint here - which is 768px, significantly wider than the 320 you were asked to look at - but the pr

Anyway, give this a try and let me know what you think or if you've got any questions about this.

allan343:
@PeterBreen thanks for the suggestions.  I made the changes you wanted.  Can you take another look?

PeterBreen (approved):
Works for me in my testing, looks good. Thanks!

PR #712, opened 2021-07-29
changed allignment of button from left to right

meganaboddam:
deleted the div wrapper around buttons "Edit Project" and "Contact Project"

### DIFF
--- a/common/components/common/projects/ContactProjectButton.jsx
+++ b/common/components/common/projects/ContactProjectButton.jsx
@@ -185,10 +185,8 @@ class ContactProjectButton extends React.PureComponent<Props, State> {
     if (this.state) {
       if (CurrentUser.isLoggedIn()) {
         return (
-          <div>
-            {this.displayEditProjectButton()}
-            {this.displayContactProjectButton()}
-          </div>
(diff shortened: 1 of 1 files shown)


common/components/common/projects/ContactProjectButton.jsx
```diff
@@ -185,10 +185,8 @@ class ContactProjectButton extends React.PureComponent<Props, State> {
     if (this.state) {
       if (CurrentUser.isLoggedIn()) {
         return (
-          <div>
-            {this.displayEditProjectButton()}
-            {this.displayContactProjectButton()}
-          </div>
+          this.displayEditProjectButton(), 
+          this.displayContactProjectButton()
```
marlonkeating, line 189:
```suggestion
          <React.Fragment>
            {this.displayEditProjectButton()} 
            {this.displayContactProjectButton()}
          </React.Fragment>
```

The render function needs to return a react element.  If you need to return multiple elements without an enclosing <div>, you can use React.Fragment.

marlonkeating (changes requested):

marlonkeating (approved):

PR #719, opened 2021-08-06
Filter malicious requests based on URL or IP address

### LINKED ISSUE #708: Filter out requests with script injection in the url
We have been probed by bots attempting javascript and sql injection attacks as part of the url.  Though our site doesn't have any known vulnerabilities to such attacks, the presence of script attacks in the url is a clear signal that the request should at the very least short-circuit without consuming unnecessary resources.

## Requirements ##
- Return error 400 to requests with javascript or sql code in the url
- Do the url check before performing any other intensive operations (such as database access)

marlonkeating:
This change adds the capability to filter out requests deemed to be malicious, due to matching patterns of known malicious URL substrings or IP addresses.

## Environment Variables ##
- MALICIOUS_URL_PATTERNS: comma-separated list of regex strings.  Any pattern matches in the url will raise an error 400
- MALICIOUS_FWD_PATTERNS: comma-separated list of regex strings.  Any pattern matches in the X-Forwarded-For header or  django's determined ip address will raise an error 400.

closes #708


justBanks (approved):
I see ... So Django disallows SuspiciousOperation by default, and the response is returned accordingly

Issue #592, opened 2021-04-18
Signing in with OAuth should return user to previous page they were browsing


marlonkeating:
When users log in from the site, they are redirected to the from which they clicked the Log In button.  However, this doesn't happen when users log in using an OAuth source like Github, LinkedIn, or Facebook.

PR #720, opened 2021-08-08
Redirecting to previous page

pravalikarouthu:
Added the logic to redirect to previous page when user signs in from social provider

https://github.com/DemocracyLab/CivicTechExchange/issues/592


marlonkeating (changes requested):
When I try to log in with github, I get the error message:
NameError at /accounts/github/login/callback/
name 'section_url' is not defined

PR #724, opened 2021-08-11
Add request throttling for anonymous and authenticated users

### LINKED ISSUE #710: Throttle requests to mitigate against bot traffic
Lately we have been having issues with bot traffic, to the extent it has caused site outages during recent weeks.  As one part of detecting and mitigating against this traffic, we would like to throttle requests that are flooding in at greater rates than we would expect from an organic user.

## Requirements ##
- Unauthenticated requests are limited to 100/min by default
- Rate limit can be configured via environment variable in order to clamp down during attack periods, or to relax/disable throttling for the purposes of performance tests.
- Throttled requests are logged

## Resources ##
https://www.django-rest-framework.org/api-guide/throttling/

marlonkeating:
This change adds a configurable rate limit to our main page view, in order to mitigate against bot traffic surges.

## Environment Variables ##
Note: Format is '#/time period'. Valid time periods include second, minute, hour or day
THROTTLE_RATE_ANONYMOUS: Max allowable rate for unauthenticated requests
THROTTLE_RATE_AUTHENTICATED: Max allowable rate for logged in users

closes #710

### DIFF
--- a/civictechprojects/views.py
+++ b/civictechprojects/views.py
@@ -39,6 +39,8 @@ from civictechprojects.helpers.context_preload import context_preload
 from common.helpers.front_end import section_url, get_page_section, get_clean_url, redirect_from_deprecated_url
 from common.helpers.redirectors import redirect_by, InvalidArgumentsRedirector, DirtyUrlsRedirector, DeprecatedUrlsRedirector
 from django.views.decorators.cache import cache_page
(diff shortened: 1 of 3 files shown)


justBanks (approved):
Implemented with just a couple decorators, very elegant

Issue #483, opened 2020-11-30
Project Profile: Separate volunteers call from project call


glenbraun:
The volunteers section of the project json is often the largest section, and it's not at the top of the page, so we should load it separately

**Checklist**

- [ ] Don't return application text for non-admin

https://trello.com/c/h3z0eZnD

PR #729, opened 2021-08-13
Add separate call for project volunteers

rohitk18:
#483

### DIFF
--- a/common/components/controllers/AboutProjectController.jsx
+++ b/common/components/controllers/AboutProjectController.jsx
@@ -35,11 +35,6 @@ class AboutProjectController extends React.PureComponent<{||}, State> {
       this.loadProjectDetails.bind(this),
       this.handleLoadProjectFailure.bind(this)
     );
-    ProjectAPIUtils.fetchProjectVolunteerList(
-      projectId,
-      this.loadProjectVolunteerList.bind(this),
-      this.handleLoadProjectVolunteersFailure.bind(this)
-    );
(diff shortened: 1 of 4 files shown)


rohitk18:
This change is for issue #483. 
A separate API endpoint for volunteers list is created and the volunteers' data in the project call is removed.
This new API call happens independently to page load so the volunteers' details will be shown when data is received.

common/components/controllers/AboutProjectController.jsx
```diff
@@ -35,6 +35,11 @@ class AboutProjectController extends React.PureComponent<{||}, State> {
       this.loadProjectDetails.bind(this),
       this.handleLoadProjectFailure.bind(this)
     );
+    ProjectAPIUtils.fetchProjectVolunteerList(
```
marlonkeating, line 38: I'm thinking we should initiate this call once the fetchProjectDetails call returns, to avoid flooding the server with too many simultaneous requests.

rohitk18: Taking your opinion into account, I have modified and committed the code to implement volunteer list api call after the project call takes place and page render completes. 
Also taking the SEO factor into account, I ensured that prerender function executes after the volunteers list loads up so that the whole page content can be grabbed and returned to web crawlers. Please check this [webpage](https://docs.prerender.io/article/25-empty-or-partially-rendered-pages) for details.

marlonkeating (approved):

PR #731, opened 2021-08-14
Changed "skills needed" to "roles needed" in different places

HebuPatil:
Changed "Skills Needed" to "Roles Needed" in the project card, project profile, and the filter name on Find Projects as per design request.


PeterBreen (approved):
Looks good 👍

PR #732, opened 2021-08-14
Updated Volunteer Approval Process Text

HebuPatil:
Updated Volunteer Approval text in email sent to project owner about a new volunteer and on the project profile page when viewed by a project owner.

### DIFF
--- a/common/components/common/volunteers/VolunteerSection.jsx
+++ b/common/components/common/volunteers/VolunteerSection.jsx
@@ -457,7 +457,7 @@ class VolunteerSection extends React.PureComponent<Props, State> {
       !_.isEmpty(pendingVolunteers)
       ? this._renderVolunteerSection(
           pendingVolunteers,
-          "Waiting for confirmation..."
+          "Review Volunteer Applicants"
         )
       : null;
   }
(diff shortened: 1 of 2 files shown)


PeterBreen (changes requested):
The changes to the VolunteerSection look good, but there's an issue with the emails.py buttons having mismatched case:

image

I checked with our design team and they want to remove ALL CAPS from all our buttons, but as you can see the prior button text was in all caps, and the design team only got this info to me yesterday -- so there's no way you could've known. Still, it means there's two ways to fix this, your choice:

1) all-caps "Review Applicant on Project Page" so it's consistent with the current capitalization method, and we'll update the text case in a separate PR.

2) Don't all-caps it, and instead change every other button text in emails.py to consistently follow the [Buttons Guide](https://docs.google.com/spreadsheets/d/153varpAD0_dD2R0rdDZ71hZqGqHO02Smu5zBfqNtslU/edit#gid=438623425) we're working on implementing. See the "Button Labels" tab, follow the "Primary CTA, Secondary, Tertiary..." rule for all of these.

Either option is fine - whether you just want to keep this PR exactly scoped to the issue at hand (thus, option 1) or if you want to go ahead and make some additional changes since you're already editing emails.py (option 2). Let me know if there's anything I can clarify.

HebuPatil:
I did option 2, please review these changes to make sure I didn't miss anything.

PeterBreen (approved):
I checked every button here and I don't see any missing. Looks great, thanks!

PR #740, opened 2021-08-25
fixed issue with clicking back from Signup/Forgot password page crash

AlenaTrushnikova:
### Issue Overview

The issue "Cannot dispatch in the middle of a dispatch" occurred because in the `chrome/MainHeader` Component the function `_handleHeightChange` included dispatcher that was called inside another dispatch and crashing the page. The issue occurred with the part of the dispatcher that was storing `headerHeight` parameter. Header height is used in the `components/common/JumpAnchor.jsx`. When one clicks the "Partner With Us" button on [Companies page](https://www.democracylab.org/companies) it jumps to the form "Take the First Step!" and offsets the Header Height. 

### Issue fixture

1. Remove `type: "SET_HEADER_HEIGHT"` for dispatcher because we use can not it this way. With this solution, there is no more need for the `PageOffsetStore` Component.
2. Pass Header Height to the `JumpAnchor` Component through the `SectionController` Component props.

### Typo

Renamed `CoroporateHackathonController` to `CorporateHackathonController` (extra letter 'o' after first 'r')


PeterBreen:
Closing due age, issue also addressed in  #971

PR #741, opened 2021-08-26
Added the logic to redirect to previous page when user signs in from …

pravalikarouthu:
Redirecting to the page from which they clicked the Login button using OAuth source.

https://github.com/DemocracyLab/CivicTechExchange/issues/592

### DIFF
--- a/common/components/common/integrations/SocialMediaSignupSection.jsx
+++ b/common/components/common/integrations/SocialMediaSignupSection.jsx
@@ -7,7 +7,7 @@ import GithubSVG from "../../svg/github.svg";
 import GoogleSVG from "../../svg/google.svg";
 import LinkedInSVG from "../../svg/linkedin.svg";
 import type { Dictionary } from "../../types/Generics.jsx";
-import url from "../../utils/url.js";
+import urlHelper from "../../utils/url.js";
 import Section from "../../enums/Section.js";
 import GlyphStyles, { Glyph, GlyphSizes } from "../../utils/glyphs.js";
 import _ from "lodash";
@@ -109,7 +109,7 @@ class SocialMediaSignupSection extends React.Component<Props, State> {
   _renderEmailSignup(): React$Node {
     return (
       <div className="LogInController-socialLink">
-        <a href={url.section(Section.SignUp)} key="email">
+        <a href={urlHelper.section(Section.SignUp)} key="email">
           <span className="LogInController-socialIcon">
             <i className={Glyph(GlyphStyles.EnvelopeSolid, GlyphSizes.X3)} />
           </span>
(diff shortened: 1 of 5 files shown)


common/components/common/integrations/SocialMediaSignupSection.jsx
```diff
@@ -124,12 +128,23 @@ class SocialMediaSignupSection extends React.Component<Props, State> {
...
+              ) : (
+              <a href={"/api/login/" + app  + "?prevPage=" + this.state.prevPage} key={app}>
                 <span style={{ color: config.iconColor }} className={iconClass}>
                   {config.iconElement}
                 </span>
                 <span>Continue with {config.displayName}</span>
               </a>
+              
+             )}
             </div>
```
marlonkeating, line 148: The only difference between the link elements in this conditional is the url.  I would set the url in a separate variable above the return statement.

Also, you can simplify constructing the url arguments with
https://github.com/DemocracyLab/CivicTechExchange/blob/52e3996e3c71941f2eb0f7d628ba285b512bc3ec/common/components/utils/url.js#L256

marlonkeating (changes requested):
Good job, the fix seems to work well!  There's just a few code cleanliness items to address.  

If you can, also run [Flow ](https://flow.org/) on your javascript, which is something we're trying to use as a team.

common/components/common/integrations/SocialMediaSignupSection.jsx
```diff
@@ -122,14 +128,23 @@ class SocialMediaSignupSection extends React.Component<Props, State> {
           const iconClass: string =
             "LogInController-socialIcon" +
             (config.iconClass ? " " + config.iconClass : "");
+
+	  var url = "";
+          if (this.state.prevPageArgs) {
+           url = "/api/login/" + app + "?prevPage=" +this.state.prevPage + "&prevPageArgs=" +    this.state.prevPageArgs 
+          } else {
+             url = "/api/login/" + app  + "?prevPage=" + this.state.prevPage
+          }
```
marlonkeating, line 137: This logic can be compressed using helper functions.
```suggestion
	  const args: Dictionary<string> = {prevPage: this.state.prevPage};
	  if(this.state.prevPageArgs) {
	    args["prevPageArgs"] = this.state.prevPageArgs;
	  }
	  const url: string =  "/api/login/" + app + urlHelper.argsString(args);
```

marlonkeating (changes requested):
I'm seeing some irregular spacing in the javascript code, which running [Flow](https://flow.org/) should solve automatically.

common/components/common/integrations/SocialMediaSignupSection.jsx
```diff
@@ -122,14 +128,22 @@ class SocialMediaSignupSection extends React.Component<Props, State> {
           const iconClass: string =
             "LogInController-socialIcon" +
             (config.iconClass ? " " + config.iconClass : "");
+
```
pravalikarouthu, line 131: Verified these changes to be working locally and hence committed it.

marlonkeating (approved):

PR #745, opened 2021-09-01
Redirect from invalid Video links

### LINKED ISSUE #742: Videos: Hard redirect to /videos/overview when invalid slug in url
When an invalid video url slug is entered, we show the overview video.  However, we should do a hard redirect to /videos/overview to make it clear that is the proper url.

marlonkeating:
- Hard redirects to /videos/overview if invalid video id given
- Fix startup crash if VIDEO_PAGES not set

resolves #742


PeterBreen (approved):
This has resisted all of my attempts to break it with bad/invalid links. Excellent 👍

PR #749, opened 2021-09-09
max height

allan343:
When a user uploads a very tall image, it no longer distorts project profile and group profile pages.


PeterBreen (changes requested):
This branch is based on a copy of master that's [quite old](https://github.com/DemocracyLab/CivicTechExchange/network); so let's get it up to date before review, please. To do that, in your dev environment:
```
git checkout master
git pull
git checkout allan_tall_images_distort_project_profile
git merge master
```
What this'll do is update your local copy of master, then merge those changes into the feature branch. That way we can do a better comparison of the fix vs the current state of dlab's site so we avoid any potential regression errors.

For example, one change not reflected here is that this rule has been moved into `_Profile.scss` as part of a refactor on the project profile pages.

That aside, I think you zeroed in on the right value to change so once you get the git issue wrestled with it ought to be pretty fast to test.

civictechprojects/static/css/partials/_Profile.scss
```diff
@@ -126,7 +126,7 @@
     flex-basis: auto;
     img {
       width: 100%;
-      max-height: unset;
+      max-height: 250px;
```
PeterBreen, line 129: This looks good, but as one last polish item -- we can just remove this line entirely. 

If you look at L21, max-height 250 is already declared so if we just remove this rule entirely it'll fall back on that and not require re-declaring it here.

image
As you can see from the element inspector, this max-height in line 129 is overriding the prior one - and since they're the same number we may as well just do it once. 

Otherwise good fix, looks like the merge worked great, ready for approval as soon as you get that in.

PeterBreen (changes requested):
One last change for cleaning up our CSS, but functionally you got it exactly right.

allan343:
Hi I made the chang requested

civictechprojects/static/css/partials/_Profile.scss
allan343: Hi I made the change requested

PeterBreen (approved):

PR #752, opened 2021-09-13
Bug Fix: Allow full clearing of tag fields

### LINKED ISSUE #751: Create Project: Cannot delete last tag in a list
For fields like Technologies Used, where one or more tags can be selected, if a user tries to delete all the tags in one of these fields, the change will fail to save.

marlonkeating:
We needed to distinguish from tag fields that were not set and tag fields that were set with a blank value, to determine whether to clear them.

resolves #751


PeterBreen (approved):
Verified the issue without the fix, tested with the fix, looks good.

PR #757, opened 2021-09-15
Uploading file over 50 characters long with no filename extension fixes

### LINKED ISSUE #668: Uploading file over 50 characters long with no filename extension fails
When uploading a file with no file extension, that is over 50 characters, saving the file fails because it treats the entire file name as a file name extension, and there are only 50 characters allowed for file name extensions.

zh-zhang916:
As the title says.

close #668

### DIFF
--- a/civictechprojects/models.py
+++ b/civictechprojects/models.py
@@ -829,10 +829,10 @@ class ProjectFile(models.Model):
         else:
             old_files = list(ProjectFile.objects.filter(file_user=owner.id, file_category=FileCategory.ETC.value)
                              .values())
-
+                             
         for file in added_files:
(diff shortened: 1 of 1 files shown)


civictechprojects/models.py
```diff
@@ -886,8 +886,12 @@ def get_owner(self):
     @staticmethod
     def from_json(owner, file_category, file_json):
         file_name_parts = file_json['fileName'].split('.')
-        file_name = "".join(file_name_parts[:-1])
```
zh-zhang916, line 889: As I talked about during the meeting, the original code is excluding the last element and including everything else into the file_name field. It will cause a problem when a file has only a filename, making file_name to be an empty string.

marlonkeating (approved):

PR #759, opened 2021-09-20
Cache user context

### LINKED ISSUE #737: Eliminate unneeded projects, groups, events calls on authenticated page views
In order to determine whether an authenticated user should see links to view their projects, groups, or events, we currently send a request to the respective endpoints on every page view.  Instead, we should flag whether the user has those entities when constructing the user context.  By doing so, we will eliminate 3 unnecessary calls with every authenticated page views.

## Checklist ##
- Cache flags of whether a user has projects, groups, and events (as this is something that should not change often)
- Inject flags into user context for each page view: https://github.com/DemocracyLab/CivicTechExchange/blob/f514e64b6d04d7f9c859309dcbb08b3ed11f07c2/civictechprojects/views.py#L404-L406
- Prevent unnecessary calls to my_projects, my_groups, and my_events that happens as part of https://github.com/DemocracyLab/CivicTechExchange/blob/f514e64b6d04d7f9c859309dcbb08b3ed11f07c2/common/components/chrome/MainHeader.jsx#L69

marlonkeating:
Currently there is a lot of user information that we use in every request, that is returned via a number of separate calls.  To optimize this, we have wrapped this information into the general page context.  To further optimize, we cache this user context so it won't need to be re-queried for 5 minutes (or if user context changes happen, whichever is sooner).

## Additional Optimizations ##
- Populate My Projects page header on back-end 
- Populate My Groups page header on back-end 

resolves #737


PeterBreen (approved):
I don't have automated tools to test this but using my browser devtools and a lot of refreshes to try and smooth out edge cases, I do see the caching is working, one call instead of three calls is working, and blocking time (as a part of page load) seems to be down -- though overall page load time does not seem to be changed much if at all. 

So in the absence of proper benchmarking (which we might want to look into) I can say the changes exist and are real, and at least help their respective parts of page load, which is what we expected anyways.






So, the optimization is there and is real, though it's not going to

PR #766, opened 2021-09-23
Favorite projects

marlonkeating:
Users can now mark favorite projects by clicking the heart icon on project cards.  They can then filter by their favorite projects on the search pages.


civictechprojects/static/css/partials/_ProjectCard.scss
```diff
@@ -98,3 +99,11 @@
   padding: 0.25rem 0.5rem;
   margin: 0.25rem 0.5rem 0.25rem 0;
 }
+
+.ProjectCard-favorite {
+  position: absolute;
+  top: 0px;
+  right: 0px;
+  padding-right: 5px;
+  padding-top: 5px;
```
PeterBreen, line 108: Instead of absolute/top 0/right 0, try
```css
float: right;
padding-top: 4px;
padding-right: 4px;
```
That should position this correctly and let text wrap underneath, though in most cases text will not wrap under the heart since it needs three lines to do so. Definitely worth testing on narrow mobile devices.

civictechprojects/static/css/partials/_ProjectCard.scss
```diff
@@ -53,6 +53,7 @@
   margin-right: 0.375rem;
 }
 .ProjectCard-title {
+  position: relative;
```
PeterBreen, line 56: Can remove this, see changes on L104-L108.

civictechprojects/static/css/partials/_ProjectFilterContainer.scss
```diff
@@ -138,3 +138,12 @@
     margin: 0;
   }
 }
+
+.favorite-filter {
+  font-weight: bold;
+  padding-left: 5px;
+
+  i {
+    color: $color-red
```
PeterBreen, line 147: Duplicate the style changes above, then add the 8px left/right margin requested - top and bottom are already handled with the parent padding so we don't need more of it here.

```css
i {
color: $color-favorite;
font-size: 24px;
margin: 0 8px;
```

common/components/componentsBySection/FindProjects/FavoriteFilter.jsx
```diff
@@ -0,0 +1,59 @@
...
+    return (
+      <div className="favorite-filter" onClick={this.doToggle.bind(this)}>
+        Favorites
+        <IconToggle
+          toggled={this.state.favoritesOnly}
+          toggleOnIcon={GlyphStyles.HeartFilled}
+          toggleOffIcon={GlyphStyles.HeartEmpty}
+          size={GlyphSizes.X2}
+        />
+      </div>
```
PeterBreen, line 54: Technically, the spec for the filter in the nav has the favorite heart always being filled in, but we then lose the visual cue that favorites are enabled as a filter. We should ask design how they want it to be made known if favorite filtering is enabled. Maybe for now something like this mockup?
image

My thought is if we're putting favorites in the filter area because it's a filter, we need to make it a filter all the way and that means having the pill show up when the favorites filter is active. What do you think?

civictechprojects/static/css/partials/_ProjectFilterContainer.scss
```diff
@@ -138,3 +138,12 @@
     margin: 0;
   }
 }
+
+.favorite-filter {
+  font-weight: bold;
+  padding-left: 5px;
```
PeterBreen, line 144: I did a few iterations on this to figure out how best to display it, heart on the left, heart on the right, and what felt best to me was Favorites on the left and then the heart icon, but not all the way over on the far right like where the dropdown chevron icons are.

Like so:
image

To do that,
```css
.favorite-filter {
  font-weight: bold;
  padding: 12px 5px;
  display: flex;
  align-items: center;
  }
```

We will also need to add .favorite-filter to the rules which also add 15px margin on mobile, or else we get this problem:
image


To do that, see here:
https://github.com/DemocracyLab/CivicTechExchange/blob/d5f9b6c6b9e76ed60c7d7805b71f9932656acecc/civictechprojects/static/css/partials/_ProjectFilterContainer.scss#L109-L112

and here:
https://github.com/DemocracyLab/CivicTechExchange/blob/d5f9b6c6b9e76ed60c7d7805b71f9932656acecc/civictechprojects/static/css/partials/_ProjectFilterContainer.scss#L135-L140

Adding favorite-filter as another class to both existing rules should do it.

PeterBreen (changes requested):
I got figma to open so I copied in some information from there to here; hoping you won't have to try and open it yourself to make the changes. Some interpretation was required so if you make a change and it looks wrong - we can discuss for sure. 

I did not (yet) follow up with design about the grey background or the question on showing whether or not the filter is active - let's make these changes first and see where we are, then loop back to them. I think it'll be easier to zero in on exactly the issues to discuss if we knock out these 'known issues' first.

civictechprojects/static/css/partials/_FavoriteToggle.scss
```diff
@@ -10,3 +9,15 @@
...
+
+.favorited {
+  color: $color-favorite;
+}
+
+.unfavorited {
+  color: $color-grey-unselected;
+}
+
+.unfavorited:hover {
```
PeterBreen, line 21:
```suggestion
.favorite-toggle:hover .unfavorited {
```

This should make the hover color change apply on the entire click target and not just the icon specifically.

PeterBreen (approved):

PR #768, opened 2021-09-27
Sprint 65 fixes

marlonkeating:
## Project Profile ##
- Fix crash on page when user not logged in

## My Events ##
- Fix: My Events link shown upon having groups, not events

## Caching ##
- Recache user context when they perform volunteering actions (or have volunteering actions performed upon them)
- volunteer_renewal_reminders: Recache user context and project when concluding volunteers when their volunteering period expires


PeterBreen (approved):
Couldn't get the site to crash on viewing pages while logged in or out, caching purges look right too.

PR #769, opened 2021-09-28
Fixed #758 Event Profile: Don't show projects section for event with no linked projects

### LINKED ISSUE #758: Event Profile: Don't show projects section for event with no linked projects
When an event has no projects associated with it, it shows the project search section with an eternal 'Loading' message at the bottom.  We should not show this section if there are no projects in the event yet.

image

wuhuwuhuhu:
closes #758

### DIFF
--- a/common/components/common/projects/ProfileProjectSearch.jsx
+++ b/common/components/common/projects/ProfileProjectSearch.jsx
@@ -14,14 +14,27 @@ type Props = {|
   wide: boolean,
 |};
 
-class ProfileProjectSearch extends React.PureComponent<Props> {
+type State = {|
+  hidden: boolean,
+|};
+
+class ProfileProjectSearch extends React.PureComponent<Props, State> {
   constructor(props: Props): void {
     super();
+    this.state = {
+      hidden: true
+    };
+  }
+
+  handleEmptyProject(): void {
+    this.setState({
+      hidden: false
+    });
   }
 
   render(): ?$React$Node {
     return (
-      <React.Fragment>
+      this.state.hidden?<React.Fragment>
         {!this.props.viewOnly && this._renderSearchControls()}
         <div className="row">
           <ProjectCardsContainer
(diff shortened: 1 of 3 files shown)


marlonkeating (changes requested):
Overall it works well!  My only critique is that it doesn't follow our readability coding conventions laid out by [Flow](https://flow.org/).

common/components/common/projects/ProfileProjectSearch.jsx
```diff
@@ -11,16 +11,29 @@ import CollapsiblePreviewPanel from "../CollapsiblePreviewPanel.jsx";
 
 type Props = {|
   viewOnly: boolean,
-  wide: boolean,
+  wide: boolean
 |};
 
-class ProfileProjectSearch extends React.PureComponent<Props> {
+type State = {|
+  hidden: boolean
```
marlonkeating, line 18: Let's either reverse the logic or pick a different name for this, as hidden gets set to false when we want to hide the section.

common/components/componentsBySection/FindProjects/ProjectCardsContainer.jsx
```diff
@@ -100,6 +103,14 @@ class ProjectCardsContainer extends React.Component<Props, State> {
   }
 
   _renderCards(): React$Node {
+    if (
+      this.state.error ||
+      (this.state.projects &&
+        this.state.projects.size === 0 &&
```
marlonkeating, line 109:
```suggestion
      (_.isEmpty(this.state.projects) &&
```

marlonkeating (changes requested):

common/components/componentsBySection/FindProjects/ProjectCardsContainer.jsx
wuhuwuhuhu: After testing, if change it like this, it will always hide the component

marlonkeating (approved):

Issue #746, opened 2021-09-02
Remove News Page and redirect to blog


marlonkeating:
All of our articles are from 2019 or earlier, and our blog is the more up-to-date news source at this point, so we should remove the news page for now.

## Checklist ##
- Remove [PressController](https://github.com/DemocracyLab/CivicTechExchange/blob/65c47dbd1c500f87b963659700b27418b0495518/common/components/controllers/PressController.jsx) component and all references
- Redirect /about/news to blog url

PR #771, opened 2021-09-29
Remove news button and redirect to blog url

zh-zhang916:
Deprecating Press section.


zh-zhang916:
As the title says.

close #746

common/helpers/front_end.py
```diff
@@ -20,6 +20,8 @@ def arg_string(idx, key, value):
 
 
 def section_url(section, args_dict=None):
+    if section == 'Blog':
+        return settings.BLOG_URL
```
marlonkeating, line 24: We don't want to put in special cases like this, otherwise if we need to do more external redirects in the future things will become quite messy.

marlonkeating (changes requested):
I'm getting an error when trying to navigate to the news page with your changes.

image

marlonkeating (approved):

PR #780, opened 2021-10-07
Added logic to for rendering emails in the background using django_rq

pravalikarouthu:
The background worker container picks the messages and prints the message on the terminal. In order to truly test this, this code has to be deployed and tested. For now, there should be no regression as the code executes smoothly without any issues.


marlonkeating:
Issue has been resolved by #812

PR #784, opened 2021-10-08
Event Profile: Enhance description readability

marlonkeating:
We want to reduce the width of the event description field on the event profile page in order to improve readability.


PeterBreen (approved):
Does exactly what design requested, can't argue with that.

PR #801, opened 2021-10-31
Throttle user creation calls to qiqochat and fix an ImportError

### LINKED ISSUE #579: OAuth: Don't publish users to Qiqochat until they have entered name
Since we have added support for letting users sign up with OAuth accounts that do not have names, we will need to make sure that we don't create accounts in Qiqochat until the user enters their name.

https://github.com/DemocracyLab/CivicTechExchange/blob/8dd76e4b687409f399ecbf8f92b368b9bb5a920d/common/helpers/qiqo_chat.py#L22-L23

zh-zhang916:
Added conditions on user name to throttle user creation calls. Did a quick fix to an ImportError found while running the unit test suite.

close #579


common/helpers/qiqo_chat.py
```diff
@@ -25,15 +25,17 @@ def __init__(self, contributor):
         # Throttle user creation calls to qiqochat
         log_result = 'Subscribe attempt to qiqochat for {user} at {time}'\
             .format(time=timezone.now(), user=self.user.id_full_name())
-        if not self.user.qiqo_signup_time \
-                or (timezone.now() - self.user.qiqo_signup_time).total_seconds() > settings.QIQO_SIGNUP_TIMEOUT_SECONDS:
+        if self.user.qiqo_signup_time \
+                and (timezone.now() - self.user.qiqo_signup_time).total_seconds() <= settings.QIQO_SIGNUP_TIMEOUT_SECONDS:
+            log_result += ':Aborting, last request < {time}s ago'.format(time=settings.QIQO_SIGNUP_TIMEOUT_SECONDS)
+        elif self.user.full_name() == ' ':
```
marlonkeating, line 31:
```suggestion
        elif self.user.full_name() == None or self.user.full_name() == "" or self.user.full_name().isspace():
```
We want to cover all the possible bases of the blank name case.  Would recommend making this into a python helper method, similar to the javascript equivalent we use: https://github.com/DemocracyLab/CivicTechExchange/blob/b7d71d0c3ea144f71f904b6a8490abe2afcf0f84/common/components/utils/string.js#L9

common/tests/test_helpers.py
```diff
@@ -1,7 +1,7 @@
 import re
 from django.conf import settings
 from django.test import TestCase
-from common.helpers.caching import is_sitemap_url
+# from common.helpers.caching import is_sitemap_url
```
marlonkeating, line 4: Can go ahead and delete line if not used

marlonkeating (changes requested):

marlonkeating (approved):

PR #803, opened 2021-11-02
[E4G-88] Adds Trello Actions to the Project Recent Activity Feed

ibnipun10:
Feature List

1. Trello model to store trello action events
2. Trello helper class to get action events from the trello api
3. Trello project updates to save to trello tables
4. Extracting last activity date from trello tables 
5. UI card container to show the changes 
6. Trello unit test for helper  ( WIP )
7. Trello command line to run trello extracts from apis to db : `python manage.py project_external_updates --trello`


civictechprojects/models.py
```diff
@@ -612,6 +619,56 @@ def to_json(self):
...
+        trello_action.id = id
+        trello_action.action_data = action_data
+
+        trello_action.save()
+
+    def get_avatar_url(self):
+        '''
+        Trello's API gives back avatar base urls (e.g. https://trello-members.s3.amazonaws.com/{id}/{hash}), but
+        a size and file extension need to be added to form a valid url
+        Currently, 30, 50, 170, & original are valid sizes
```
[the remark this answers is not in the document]
ldahill: There are really 2 avatar URLs that Trello uses - a public url and a non-public url. 
We will be retrieving the non-public url, which can be different from the non-public url used within the Trello app.
In my case the non-public url displays my profile picture and the public url displays my initials (per my Atlassian account privacy permissions)

common/management/commands/project_external_updates.py
```diff
@@ -21,6 +57,87 @@ def handle(self, *args, **options):
...
                 pass
 
+def get_trello_board_id_from_url(url):
+    board_id = None
+    if url is not None:
+        # it will be in the format http[s]://trello.com/b/boardid/....
+        token = url.split('/')
+        
+        #board id will be at 4th position
+        #TODO : Check the URL using regex sample regex : http.://trello.com/b/(.*)[/]*.*
```
ldahill, line 67: Does this still need to happen?

civictechprojects/models.py
```diff
@@ -612,6 +619,56 @@ def to_json(self):
...
+class TrelloAction(models.Model):
+    action_project = models.ForeignKey(
+        Project, related_name='trello_actions', on_delete=models.CASCADE)
+    member_id = models.CharField(max_length=2083)  # action creator
+    member_fullname = models.CharField(max_length=200)
+    member_avatar_base_url = models.CharField(max_length=2083, blank=True)
+    board_id = models.CharField(max_length=2083)
+    action_type = models.CharField(max_length=2083)
+    action_date = models.DateTimeField(auto_now=False)
+    id = models.CharField(max_length=2083, primary_key = True)
```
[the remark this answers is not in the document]
ibnipun10: Sorted.... changed the length of the fields.

common/management/commands/project_external_updates.py
ibnipun10: No need to use regex as board id always exists at this position.

common/management/commands/project_external_updates.py
```diff
@@ -21,6 +49,89 @@ def handle(self, *args, **options):
...
+    """
+    Queries for Trello Board ProjectLinks. Selects one Trello link per project,
+    prioritizing links with 'link_projmanage' as the name.
+    """
+    from civictechprojects.models import ProjectLink
+    return ProjectLink.objects.filter(link_url__icontains='trello.com/b')\
+        .exclude(link_project__isnull=True)\
+        .extra(select={'is_projmanage_link': "link_name = 'link_projmanage'"})\
+        .order_by('link_project_id', '-is_projmanage_link')\
+        .distinct('link_project_id')
```
marlonkeating, line 133: With github, we support ingesting multiple github repos per project, so we should do the same for trello.

marlonkeating (changes requested):

common/management/commands/project_external_updates.py
ldahill: Pushed an update for this.
This should now support multiple Trello board links

marlonkeating (approved):

PR #806, opened 2021-11-04
Live Event: Allow admins to impersonate qiqochat users

### LINKED ISSUE #805: Let admins log into Qiqochat embedded view as other users
To track down some issues with the integration between DemocracyLab and Qiqochat, we want to enable admins to view the Live Event page while signed in to Qiqochat as another user.

## Acceptance Criteria ##
- Impersonation only works when environment variable QIQO_IMPERSONATION_ENABLED=True

marlonkeating:
When QIQO_IMPERSONATION_ENABLED=True, users can add uuid and qiqo_uuid to the Live Event page url to impersonate another user inside the qiqochat iframe.

resolves #805


PeterBreen (approved):
We discussed the pro/cons in standup, it looks fine

PR #812, opened 2021-11-19
Offload emails to worker thread with redis queue

### LINKED ISSUE #721: Emails should be outsourced to external job
Emails sometimes take a long time to send, so we should send them to a separate job dedicated to sending emails, that won't block the request's processing.

## Requirements ##
- Calls to send emails sends email information to a queue that will be processed by a separate job
- Processing job has configurable rate limit 

## Resources ##
[Background Tasks in Python](https://devcenter.heroku.com/articles/python-rq)

marlonkeating:
This change uses a redis-powered job queue to send out emails in an asynchronous manner.

## Redis Queue ##
Add /django-rq/ to the domain url

## Environment Variables ##
REDISTOGO_URL: Url pointing to redis instance (does not need to specifically be a [Redis To Go](https://redistogo.com/) instance)
REDIS_ENABLED: When True, use the redis queue.  If false or not set, use an asynchronous thread to send the email.

closes #721


PeterBreen (approved):
This looks pretty good from my testing; added some jobs to queue and then the worker picked them up and finished them. 

the only question is whether every deployment will require setting `heroku scale worker=1` but that's a documentation question, not implementation.

PR #824, opened 2021-12-08
MyProjectsCard Bootstrap CSS

rohitk18:
Used Bootstrap CSS and removed table html.
Implemented for cards in MyProjects and MyGroups (uses same CSS classes)

### DIFF
--- a/civictechprojects/static/css/partials/_MyProjectCard.scss
+++ b/civictechprojects/static/css/partials/_MyProjectCard.scss
@@ -16,9 +16,6 @@
 
 .MyProjectCard-projectName {
   font-weight: bold;
-  &.text-overflow-break {
-    overflow-wrap: break-word;
-  }
 }
 
 .MyProjectCard-button {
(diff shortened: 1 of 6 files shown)


civictechprojects/static/css/partials/_MyProjectCard.scss
```diff
@@ -16,14 +16,9 @@
...
   font-weight: bold;
-}
-
-.MyProjectCard-table {
-  background-color: $color-background-light;
-  border: solid 1px $color-grey-frame-border;
-  margin: 20px auto;
-  padding: 16px;
-  width: 98%;
+  &.text-overflow-break {
```
PeterBreen, line 19: This works - but to save a few lines of CSS how about using the built-in bootstrap utility class `text-break` to do the same thing? ([ref](https://getbootstrap.com/docs/4.4/utilities/text/#word-break))

PeterBreen (changes requested):
This looks really good; I have one small requested change to a utility class, beyond that no other changes needed. Thanks for taking this issue! 👍

civictechprojects/static/css/partials/_MyProjectCard.scss
rohitk18: Replaced with the built-in bootstrap utility class `text-break`

rohitk18:
Replaced with the built-in bootstrap utility class `text-break` in project card and group card
Removed `text-overflow-break` from _MyProjectCard.scss

PeterBreen (approved):
👍 Perfect, thanks!

PR #826, opened 2021-12-13
Bug Fix: My Project page crashes when not logged in

### LINKED ISSUE #825: My Projects: Shows blank page when navigating to while not logged in
Navigating to the My Projects page url while not logged in shows a blank page.  This is a problem, since some of our emails around volunteer renewal link to this page, and given that the volunteer in question may not have logged in for some time, they will see the blank page and be very confused.

What we need to do in this scenario is prompt the user to log in, and then immediately redirect them to the My Projects page.

marlonkeating:
resolves #825


PeterBreen (approved):
Compared this to latest master and the login redirect works, nice and straightforward.

Issue #822, opened 2021-12-03
Companies Page - style and content revision


PeterBreen:
As part of a [redesigned companies page](https://trello.com/c/Gk7d4ya2/297-website-companies-page-update-content), design has requested text changes throughout the entire page as well as a consistency pass on spacing between/within page elements. This item, therefore, has two parts:

 - Copy changes throughout the entire page, [documented here.](https://docs.google.com/document/d/1R1_2hUmc_0fU-SzIF1utHgzffOz4IGrWXwopVoPdgVk/edit#heading=h.t90uf38k145w) Note wording, capitalization, and also font changes. 
   -  A note on fonts: the document uses the semantic names from the design system, but for our purposes "Subtitle 1" is by default `<h3>` or `.subtitle1` so you shouldn't have to declare any new font sizes, just use the defaults and classes added in #618. Any questions, ping me on slack.
   - Also, any reference to "Body 2" should be ignored; just use default body text (`<p>`) instead. That doc was written before Body 2 was removed from the design system.
 
 - Comprehensive spacing, color, and layout changes for mobile and then desktop, [as documented on figma](https://www.figma.com/file/WADcmVjJh5ARVoZ09xlpfdFN/DemocracyLab?node-id=27336%3A22519). 
 
 Please be aware of #819, #820, and #821 as they all have impact on the companies page. None of them should prevent work on this issue, but they're context.

The trello/figma has some notes about updating which testimonials display on page, but for now ignore that part of the item. It needs further info from product team (or ...someone) and can be done in a separate PR.

PeterBreen:
Done in #827, closing. Design has indicated some further changes they'd like to make which will be handled in another issue.

Issue #820, opened 2021-12-02
Companies page - update contact form


PeterBreen:
As part of a [redesigned companies page](https://trello.com/c/Gk7d4ya2/297-website-companies-page-update-content), design has requested some changes to the contact form on the Companies page. We have a reusable ContactForm component which will need some modification alongside the Companies page also needing some changes.

Currently the form is called once on the page and renders below the tabbed content sections, but design would like us to move it into the tabbed sections and have section-specific header text. See the [design on figma for specifics](https://www.figma.com/file/WADcmVjJh5ARVoZ09xlpfdFN/DemocracyLab?node-id=27336%3A22519), but broadly:

 - The form should be one column, full width, grey background on mobile
 - The form should be moved from the bottom of the page and above the "What People Are Saying" section (which will be retitled "What our partners are saying" in another PR) in both the Hackathon and Sponsorship sections, with appropriate header text.
 - The form should have a maximum width of 560px, white background (with the remaining section bg as grey) on desktop
 - The "Areas of Interest" checkbox selector should be removed. However, the Hackathon section contact form should pass `interest_hackathon` to the Contact Form and the Sponsorship section contact form should pass `interest_sponsorship` so that the email generated will still let the recipient know contextual information about the message. Basically, these should be invisible to the end user but 'pre-checked' for the form submit.
 - any changes to ContactForm should not affect how the form operates on other pages, such as the Contact Us page.

relevant files:  
- https://github.com/DemocracyLab/CivicTechExchange/blob/master/common/components/controllers/CorporateHackathonController.jsx
 - https://github.com/DemocracyLab/CivicTechExchange/blob/master/common/components/forms/ContactForm.jsx

PeterBreen:
Done in #827, closing. Please see #821 for future contact form changes that are still open.

Issue #819, opened 2021-12-02
Companies Page - remove Impact Stories


PeterBreen:
As part of a [redesigned companies page](https://trello.com/c/Gk7d4ya2/297-website-companies-page-update-content), we are removing the "Impact Stories" section from both the Hackathon and Sponsorship portions of the page.

- remove the entire Impact Stories section from both sections of the page
- remove the BlogCarousel import statement from the page (don't delete the component, we just don't need to import it here anymore)
- remove `corporate-hackathon-stories` and `corporate-sponsorship-impact` style declarations from the sass partial for this page

relevant files:
 - https://github.com/DemocracyLab/CivicTechExchange/blob/master/common/components/controllers/CorporateHackathonController.jsx (Impact Stories sections start on L269 and L416)
 - https://github.com/DemocracyLab/CivicTechExchange/blob/master/civictechprojects/static/css/partials/_CorporateHackathon.scss

ParkerM2:
I'm interested in working on this Peter.

PeterBreen:
Handling this in the same branch as #822.

PeterBreen:
Done in #827, closing.

PR #827, opened 2021-12-14
Companies Page Update

PeterBreen:
Addresses #822, #820, #819

### to do (MVP)
- [x] #822
- [x] #820, both style and the props change
- [x] #819 

#### to-do (bring these back to design)
- [x] get magnifying glass SVG and replace the megaphone icon in Hackathon tab, "How it works" section.
- [x] for citations (Deloitte, etc) - are these linked? How? If not, where are the sources?
- [x] design review and fixes
  - [x] see [handoff updates and notes on trello](https://trello.com/c/Gk7d4ya2/297-website-companies-page-update-content); has detailed changes per section
  - [x] update vector graphics with new files from Luba
  - [x] remove bottom image with gradient fade
  - [x] remove Sponsor footer test 
  - [x] H1-H4 font color change (this page only until we get that part of the DS in)

Note: 08f81ef is a global change to our H1-H4s so we should do a spot check on all pages. As I read it, this was an intended piece of the design system that didn't work in #624 and I missed it on review. It's also frustratingly difficult for me to change only on this page so I went ahead and did it sitewide. It'll make things look different but hopefully not broken!

#### references:
 - [figma](https://www.figma.com/file/WADcmVjJh5ARVoZ09xlpfdFN/DemocracyLab?node-id=27336%3A22455)(original design)
 - [trello card](https://trello.com/c/Gk7d4ya2/297-website-companies-page-update-content) (has changes which supersede figma)


common/components/controllers/CorporateHackathonController.jsx
```diff
@@ -298,180 +289,228 @@ class CorporateHackathonController extends React.PureComponent<{||}, State> {
...
-                  responsibility.
+                  Supporting DemocracyLab shows your company’s commitment to
+                  social responsibility.
                 </p>
                 <p>
-                  Our community of volunteers is tomorrow’s diverse, dynamic,
-                  and purpose-driven workforce. Sponsorship lets them know your
-                  company helped create the opportunities they are enjoying.
+                  79% of employees prefer to work at a socially responsible
+                  company. (Deloitte)
```
marlonkeating, line 314: Should Deloitte be a link?

common/components/controllers/CorporateHackathonController.jsx
```diff
@@ -298,180 +289,228 @@ class CorporateHackathonController extends React.PureComponent<{||}, State> {
...
+          <h4>
+            Not ready yet? Learn more:{" "}
+            <a
+              href={cdn.document(
+                "2021+DemocracyLab+Corporate+Hackathon+Prospectus.pdf"
+              )}
+            >
+              Corporate Tech-for-Good Hackathons PDF{" "}
+              <i className={Glyph(GlyphStyles.PDF, GlyphSizes.X1)}></i>
+            </a>
```
marlonkeating, line 459: This is duplicated at line 402, can we break this out into a shared block?

marlonkeating (changes requested):
I'm seeing a couple irregularities on simulated 320px resolutions.

Chrome 320px:
image


Firefox 320px:
image

common/components/controllers/CorporateHackathonController.jsx
PeterBreen: Makes sense. Since they're both reused, I broke out both the Hackathon and Sponsorship prospectus PDFs into shared blocks. (Also this helped me catch a mistaken use of Hackathon in the Sponsorship tab, so two for one!)

common/components/controllers/CorporateHackathonController.jsx
PeterBreen: Yes, good catch, thanks.

PeterBreen:
> I'm seeing a couple irregularities on simulated 320px resolutions.

I saw those too, and I think I have a fix in for them. Try the latest revision and see if it works / make sure it doesn't regress elsewhere. Something in the tab content section was pushing content to a minimum width of 334px (hence why this only showed up at viewports like 320) but I put in a rule that ought to handle that.

marlonkeating (approved):

PR #829, opened 2021-12-21
Sprint 71 Fixes

### LINKED ISSUE #828: Trello updates project timestamp to now() instead of the last trello action
When a project has a trello board listed, and we pull actions from it during the project_external_updates job, the project timestamp is always updated to now() instead of the latest trello action's timestamp.

marlonkeating:
resolves #828


PeterBreen (approved):
Looks good.

PR #832, opened 2022-01-12
Contact Volunteers: Don't cc project owners on every email

### LINKED ISSUE #831: Contact Projects: Project leaders CC-ed on every email
When contacting all volunteers, the project owner/co-owner are cc-ed on all of the emails sent out to individual team members.  This results in a lot of spam for project owners, so we should not cc them.

marlonkeating:
resolves #831


PeterBreen (approved):
Looks pretty straightforward, good fix.

PR #836, opened 2022-01-17
fix ceewa30: LoginChange

ceewa30:


ceewa30:
When a user goes to log in. "Looks Good" will appear below the entry boxes - fixed

marlonkeating (changes requested):
Why was example.env removed?

ceewa30:
i renamed the file in my local

Thank You
Sivakumar

PeterBreen:
Closing due to age, #970

PR #857, opened 2022-02-23
csp_policy_update

harmeet9307:
updated csp policy to include heap analytics.


marlonkeating (changes requested):
The correct string to use is "*.heapanalytics.com/", which ensures we catch all subdomains of heapanalytics.com.

marlonkeating:
We're no longer using heap, so closing for now

Issue #856, opened 2022-02-20
Dismiss Volunteer: Update email text


marlonkeating:
We want to change the text for emails sent to volunteers when they have been dismissed from a project.

## Subject Line ##
"You have been dismissed from DemocracyLab"
Changes to
"Thank you for your work at [Project Name]."

## Email Body ##
Change email body to:
Thank you for contributing to [Project Name]. Your volunteer work for the project has ended. "[Message from project owner]" 

## Code Location ##
https://github.com/DemocracyLab/CivicTechExchange/blob/23eb1fbd253d858098f05ef713910fdbddcc804b/civictechprojects/views.py#L1038

PR #859, opened 2022-03-04
Update to email subject and body

joelchooks:
This fix updates the subject and body of email sent to a dismissed volunteer according to issue #856


PeterBreen (changes requested):
The updates to views.py for the email looks good, but there's what seems like empty changes to example.env also? If you could remove that file from the PR I think that's the only change needed.

PeterBreen:
Not sure how this got missed, but it's been solved so I'm closing it due to age.

PR #865, opened 2022-03-25
Volunteer RSVP for Event 

marlonkeating:


common/components/componentsBySection/FindProjects/ProjectCard.jsx
```diff
@@ -73,6 +73,7 @@ class ProjectCard extends React.PureComponent<Props, State> {
           {this._renderTitleAndIssue()}
           {this._renderProjectDescription()}
           {this._renderSkillsNeeded()}
+          {this._renderCardOperation()}
```
PeterBreen, line 76: Making a note for review: I need to test the project cards on homepage/about projects page due to this change, it's possible the conditional rendering needs to be here due to moving renderCardOperation up here.

common/components/componentsBySection/FindProjects/ProjectCard.jsx
```diff
@@ -169,6 +175,32 @@ class ProjectCard extends React.PureComponent<Props, State> {
...
+        buttonConfig = {
+          onClick: () => cardOperation.operation(),
+        };
+      } else {
+        buttonConfig = { href: cardOperation.url };
+      }
+      return (
+        <React.Fragment>
+          <Button
+            variant="primary"
```
PeterBreen, line 193: For the design as given, toggling between `variant="primary"` and then `variant="outline-secondary"` matches the requested look.

civictechprojects/static/css/partials/_ProjectCard.scss
```diff
@@ -103,4 +103,18 @@
...
   padding-right: 4px;
+}
+.ProjectCard-operation {
+  grid-row: 5;
+  grid-column: 1/5;
+  border-top: 1px solid $color-grey-medium;
+  text-align: center;
+}
+//selector specificity here is to avoid  matching .ProjectCard-root a, which is hilarious but not intended behavior
+a.ProjectCard-rsvp-btn, button.ProjectCard-rsvp-btn {
```
PeterBreen, line 114: It might be cleaner, longer term, to change L17 to something like  `.ProjectCard-root > a` so it only matches the anchor we want it to match, rather than having to define this level of specificity for any other links we put on a card.

PeterBreen (approved):
It works! I don't see any issues we didn't already have in our "improve this in the future" list, though I left a couple notes for later and we do still have some TODOs in text to handle later.

PR #868, opened 2022-04-12
Live Event V2

marlonkeating:


civictechprojects/static/css/partials/_AboutProjectEventDisplay.scss
```diff
@@ -18,8 +18,16 @@
     button,
     a {
       width: 100%;
+      margin: 1rem 0;
```
PeterBreen, line 21: note to self; remove this and use `gap: 1rem;` or whatever value on parent flex container (Profile-top-interactions)

common/components/common/event_projects/JoinConferenceButton.jsx
```diff
@@ -0,0 +1,45 @@
...
+    super(props);
+  }
+
+  render(): React$Node {
+    return (
+      <React.Fragment>
+        <Button
+          variant={this.props.variant}
+          type="button"
+          className={this.props.className}
```
PeterBreen, line 30: Not for this release but let's mark a todo here -- I should move the CSS that handles this button's internal layout (currently `.AboutEvent-livebutton` and `.AboutEvent-usercount`) into a specific JoinConferenceButton class/classes, and we use that here alongside any classes passed from props.

common/components/common/event_projects/JoinConferenceButton.jsx
```diff
@@ -0,0 +1,45 @@
...
+class JoinConferenceButton extends React.PureComponent<Props> {
+  constructor(props: Props): void {
+    super(props);
+  }
+
+  render(): React$Node {
+    return (
+      <React.Fragment>
+        <Button
+          variant={this.props.variant}
```
PeterBreen, line 28: One thing I didn't address was where in the component chain we want to have the safety net for variant -- the `{variant: this.props.variant || default }` -- I think that's done higher up than here right now but for code cleanliness, would it be worth bringing that statement down to this component once and not doing it higher up repeatedly?

PeterBreen (approved):
Overall, I think this is good - I hope so considering we're using it live :joy: but there are definitely some clean-up tasks we could do to help streamline the new components and to unwind some the last-minute "it works; ship it" style/code decisions -- but it works, and that's good.

common/components/common/event_projects/JoinConferenceButton.jsx
PeterBreen: I think it'll work best if I take care of this in #870 after this gets merged to master. Make sense to you?

common/components/common/event_projects/JoinConferenceButton.jsx
marlonkeating: That scenario should be handled by the buttonConfig prop, which overrides the variant if called for.

Issue #687, opened 2021-07-09
Remove "Sign in to ..." button conditional text


PeterBreen:
❗  Pre-requisite for merge (unrelated to this task's specific changes, but required before we can merge it for technical reasons): #720 

## Issue

We have buttons which conditionally change text based on a user's logged in status. For example, when a user is logged out and views a project profile:
![image](https://user-images.githubusercontent.com/16870466/125122473-56304c00-e0aa-11eb-89e1-4cb8188aec82.png)

And when logged in, viewing the same project profile:
![image](https://user-images.githubusercontent.com/16870466/125122552-706a2a00-e0aa-11eb-85aa-c35b40f2bd5d.png)

## Solution

Design has asked us to make it so buttons will no longer conditionally change their text based on login state. They should always read "Contact Project" (specifically, always read whatever the "logged in" version of the button text is, at all times) but without removing the functionality of routing a user to login if needed.

As an example,
https://github.com/DemocracyLab/CivicTechExchange/blob/b28832745a15fc40ebaf5b44a62e2ca9ca086184/common/components/common/projects/ContactProjectButton.jsx#L146-L148

This button would need to be changed to read "Contact Project" - which may also suggest refactoring the entire ContactProjectButton component so we don't repeat so much code. 

A (partial?) list of pages where we do this is:

**Project Profile**
 - Contact project 
 - Volunteer with project
 - Volunteer for a specific position

**Group Profile**
 - Contact group

**Event Profile**
 - Join Event

wuhuwuhuhu:
Please forgive me if my understanding is not correct. If we just need to change the text of these "sign to ... " buttons to the "logged in" version but still retain its functionality and redirect. Could we just change the innerHTML of the Buttons? for example:

- for the ContactProjectButton component, using "Contact Project " to substitute "Sign in to Contact Project" at line 147,
https://github.com/DemocracyLab/CivicTechExchange/blob/b28832745a15fc40ebaf5b44a62e2ca9ca086184/common/components/common/projects/ContactProjectButton.jsx#L146-L148
- for the AboutEventDisplay component, using "Join Event" to substituted "Log In to Join Event" at line 20
https://github.com/DemocracyLab/CivicTechExchange/blob/877652907422b12e934361f3e82634b1afd91345/common/components/componentsBySection/CreateEvent/AboutEventDisplay.jsx#L201

PeterBreen:
[...]

PR #873, opened 2022-05-11
Remove Sign in to ... button conditional text

Meena-Bharani:
Remove "Sign in to ..." button conditional text](https://github.com/DemocracyLab/CivicTechExchange/issues/687#)
#687


PeterBreen (approved):
Looks good, thanks for taking care of this!

PR #874, opened 2022-05-12
Sprint 78 fixes

PeterBreen:


civictechprojects/static/css/partials/_Global.scss
```diff
@@ -122,12 +116,18 @@ button:disabled,
 .btn-outline-ghost:active {
   color: $color-neutral-100;
 }
+button:disabled,
+.btn:disabled {
+  background-color: $color-neutral-25 !important;
+  border-color: $color-neutral-35 !important;
+  color: $color-text-disabled !important;
+}
```
PeterBreen, line 124: By moving the disabled state rules below all the other btn rules, they make sure disabled 'wins out' over any other styling when it comes to dueling !important rules, but I'd prefer to re-do this later and remove all those !importants. they're hacks and they work, but they cause issues like this. And this is a hack fixing a hack. It's not great!

marlonkeating (approved):

PR #875, opened 2022-05-25
PR#870 Refactor Fixes

PeterBreen:
Code refactoring and cleanup, primarily:
 - removes the dueling !important rules on disabled state buttons by refactoring the rules to cleave a little closer to Bootstrap's initial generation in `civictechprojects\statics\css\vendor\bootstrap\_buttons.scss`. 
 - removes some legacy SASS variables used in SplashScreen, but the end result should be the same
 - removes some legacy button sizing rules in SplashScreen so they conform to the current ruleset
 - fixes some legacy CSS/type declarations on Contact Group buttons

### DIFF
--- a/civictechprojects/static/css/partials/_SplashScreen.scss
+++ b/civictechprojects/static/css/partials/_SplashScreen.scss
@@ -30,8 +30,6 @@
   border: 1px solid $color-orange;
   color: $color-text-dark;
   background-color: $color-orange;
-  border-radius: 5px;
-  padding: 6px 12px;
   min-width: 80px;
 }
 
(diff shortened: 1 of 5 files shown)


civictechprojects/static/css/partials/_SplashScreen.scss
```diff
@@ -11,7 +11,7 @@
 
 .SplashScreen-fill {
   min-height: calc(
-    100vh - #{$header-height} - #{$footer-height}
+    100vh - 100px
```
PeterBreen, line 14: This is a bit out of scope for the PR but since I was in SplashScreen anyway I searched the code, saw we had two identical variables used nowhere else, and simplified. 

I expect when the nav changes go in we'll want to revist this number (some rough tests suggest the lower the better, down to a point -- say, 30-50px) but for now I just cleaned up the math, no changes in end result.

marlonkeating (approved):

PR #878, opened 2022-05-31
Library Upgrades Q2 2022

marlonkeating:
closes https://github.com/DemocracyLab/CivicTechExchange/pull/845


PeterBreen (approved):
Builds fine, no issues that I could see, and it passed testing. One future enhancement will be the sass validator to fix the division indicator, but that's best done separately.

PR #880, opened 2022-06-13
Single row nav at desktop resolution

PeterBreen:
Changes the desktop site navigation so that, when space allows, the nav will switch to a single row layout rather than two-row and otherwise operate unchanged. 

Generally, this means changes should only be seen at ~1200px or greater depending on user name length and logged in state.


marlonkeating (approved):
Changes look good, with smooth transition between one and two rows.

PR #887, opened 2022-07-13
Add private slug support for projects/groups

marlonkeating:
This change adds support for marking projects and groups as private to those who do not know the unique slug of the project/group.


PeterBreen (changes requested):
Works for admins as expected, but the slug field could use some safeguards. I tried a slug of `slug slug` and the space resulted in being unable to load the project, trying both `projects\slug slug` and encoding the space as `\projects\slug%20slug` in the url.

Seems plausible to do a string replace with a regex to just convert all spaces and special characters to dashes? That's fairly common for slug generators as I understand it.

(for bonus points, try putting in `slug%20slug` as your slug and watch django toss out a REST error page I've never seen before. To the point that I think I broke project 1030 as it won't let me save any updates.)

PeterBreen (approved):

PR #890, opened 2022-07-22
Use REDIS_URL environment variable instead of REDISTOGO_URL

marlonkeating:
We're moving away from the Redis-to-go service, and as part of that we're changing the redis url environment variable to something more generic. 

Quick update to test fixtures for including testimonials as well.


PeterBreen (approved):
env change works, testdata works 👍

Issue #709, opened 2021-07-28
Recache requests should be outsourced to external job


marlonkeating:
When a heavily-linked entity is updated, triggering a recache among multiple entities, that creates a large cascade of database operations.  While we want those linked entities to be updated within minutes of the update operation, we don't need to block the edit operation.  This becomes a particular issue when performing large batch operations such as updating projects based on github commits, which can trigger linked entity re-caches for multiple active projects.

## Requirements ##
- Recaching linked entities will now add the requeuing of each entity to a queue that will be processed later
- Queued recaching jobs will have timestamp that will signal to skip the recaching if the entity's last update date is after the requeue timestamp
- Processing job has configurable rate limit 

## Resources ##
[Background Tasks in Python](https://devcenter.heroku.com/articles/python-rq)

qchiujunhao:
Should we close this issue? @marlonkeating

PR #891, opened 2022-07-27
Salesforce

justBanks:


marlonkeating (changes requested):
The biggest issue I'm seeing right now is that we're now sending out potentially multiple Salesforce calls with each step in the project creation process.  This is already something of an issue with our recaching (https://github.com/DemocracyLab/CivicTechExchange/issues/709), so it would be best not to to add to that overhead.  Can we move the salesforce operations to the redis cache runner, similar to what we're doing with emails?

salesforce/volunteer_job.py
```diff
@@ -42,6 +39,12 @@ def save(project_position):
         #thread.start()
 
 
+def save(project_position):
+    from civictechprojects.models import Project
+    if Project.objects.get(id__exact=project_position.position_project.id).is_searchable:
```
marlonkeating, line 44:
```suggestion
    if project_position.position_project.is_searchable:
```
Does this not work?

justBanks: That way is much better. Checked it in.

salesforce/volunteer_job.py
```diff
@@ -1,46 +1,56 @@
...
+            method="PATCH",
+            url=f'{client.job_endpoint}/platform_id__c/{platform_id__c}',
+            data=json.dumps(data)
+        )
+        ''' Changed this to a synchronous call to avoid record locks (duplicate position names are possible) '''
+        SalesforceClient().send(req)
+
+        #thread = threading.Thread(target=run, args=(req,))
+        #thread.daemon = True
+        #thread.start()
```
marlonkeating, line 39: Can delete

marlonkeating (changes requested):

civictechprojects/models.py
```diff
@@ -1326,6 +1330,10 @@ def is_up_for_renewal(self, now=None):
         return self.is_approved and (self.projected_end_date - now) < settings.VOLUNTEER_REMINDER_OVERALL_PERIOD
 
 
+    def salesforce_job_id(self):
+        role = Tag.tags_field_descriptions(self.role)
+        return f'{self.project.id}{role.lower().replace(" ", "")}'
```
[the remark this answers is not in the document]
justBanks: Subtle differences between the two methods, on different models. I went ahead and used the second one in volunteer_job, which was my intent

salesforce/contact.py
```diff
@@ -3,27 +3,41 @@
...
+        data['firstname'] = contributor.first_name
+    if contributor.last_name:
+        data['lastname'] = contributor.last_name
+    if contributor.phone_primary:
+        data['phone'] = contributor.phone_primary
+    if contributor.postal_code:
+        data['mailingpostalcode'] = contributor.postal_code
+    if contributor.country:
+        data['mailingcountry'] = contributor.country
+    if contributor.date_joined.strftime('%Y-%m-%d'):
```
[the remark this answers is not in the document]
justBanks: Added the DateTimeFormat and resolved all issues

justBanks:
Seems this can't be merged until "requested changes are addressed," even though I have addressed them. Anything else I can do?

PeterBreen:
> Seems this can't be merged until "requested changes are addressed," even though I have addressed them. Anything else I can do?

Marlon has to approve the changes since he requested them, I believe. I know he's busy this morning but I don't know his schedule past that.

salesforce/volunteer_hours.py
```diff
@@ -1,7 +1,120 @@
...
+        "GW_Volunteers__Contact__r":
+        {
+            "platform_id__c": volunteer.volunteer_id
+        },
+        "GW_Volunteers__Volunteer_Job__r":
+        {
+            "platform_id__c": volunteer.salesforce_job_id()
+        },
+        "GW_Volunteers__Status__c": 'Application Received',
+        "GW_Volunteers__Start_Date__c": volunteer.application_date.strftime("%Y-%m-%d"),
```
marlonkeating, line 40: Use SALESFORCE_DATE in this file

salesforce/volunteer_job.py
```diff
@@ -1,46 +1,52 @@
...
+            "name": position_role,
+            "gw_volunteers__description__c": project_position.position_description
+        }
+        req = requests.Request(
+            method="PATCH",
+            url=f'{client.job_endpoint}/platform_id__c/{platform_id__c}',
+            data=json.dumps(data)
+        )
+        ''' Changed this to a synchronous call to avoid record locks (duplicate position names are possible) '''
+        SalesforceClient().send(req)
```
marlonkeating, line 35: Why is this particular save operation a problem to do asynchronously?

marlonkeating (changes requested):

salesforce/volunteer_job.py
justBanks: Take for example https://www.democracylab.org/projects/1 : Full Stack Developer is listed multiple times. My code comes along and picks up both of these during `merge_chanes()` and we have two updates to the same volunteer_job happening simultaneously--the error is that record locking fails.

marlonkeating: Is it possible to give it a unique name?  Like "<position name>_<position id>"

salesforce/volunteer_hours.py
justBanks: made revision and tested on dev site

salesforce/volunteer_job.py
justBanks: I believe the Full Stack Developer listings arise from tags so I don't see how to extract position ids

marlonkeating: If the project_position argument is a [ProjectPosition](https://github.com/DemocracyLab/CivicTechExchange/blob/a39efa91c0ff076513e920abe8a4286b814d61a6/civictechprojects/models.py#L992) object, it should have an id attribute you can use.

justBanks: back to asynchronous save with commit 4a8c88ac0667c4571ed90ba4efcdae9cdd92dad3

PR #893, opened 2022-07-28
Support markdown in project, event project, and group description fields

marlonkeating:


PeterBreen (changes requested):
Aside from the nesting <p> tags in the line comment, the only other issue I see is the negative margin on list items leading to some issues on group, project, and event project pages if they're the first child inside the markdown-container div, examples below.

image image

It happens for unordered and ordered lists, and does not happen if the lists are the second child element (e.g. a paragraph then a list) so, what should fix it is this:
```css
.markdown-container ol:first-child, .markdown-container ul:first-child {
  padding-top: 16px;
} 
```

This should only match if a list is the first child of a markdown-container and apply padding, else it'll do nothing. Best I can tell, that is what we want.

PeterBreen (approved):

PR #898, opened 2022-08-13
Profile link regexes #652

### LINKED ISSUE #652: Profile Links: Link source detection regexes too broad
This expression and the other social ones below are a little too generous in matching; for example `https://duckduckgo.com/?q=facebook.com` will result in a Facebook icon on the project profile, despite going to an entirely different site. Not blocking on this PR, I don't think, but we may want an engineering excellence ticket on firming up our autodetect regexes?

_Originally posted by @PeterBreen in https://github.com/DemocracyLab/CivicTechExchange/pull/647#discussion_r648521484_

schaferyan:
**LINK SOURCE DETECTION REGEXES:** Social media logos now only render if the platform name immediately follows _"http://"_, _"https://"_ and an optional _"www."_ I also replaced the Regexp constructor syntax with regex literal syntax. According to Mozilla, literals compile when the script loads rather than at runtime, and are therefore more efficient assuming you don't expect the regexes to be altered. 

**OTHER NOTES ON LINK DETECTION BEHAVIOR:** The default social media links for new projects (Facebook, Twitter, and LinkedIn) display logos by default and ignore regexes. This means that if you type any random url they will still display the respective logos. Also it appears that depending on whether _"http"_ or _"https"_ is used, these links are displayed as just the name, or with a  _".com"_. For example, if _"http://facebook.com"_ is supplied, the name of the link appears as _"Facebook"_, if _"https://facebook.com"_ is used, it will display as _"Facebook.com"_.

Closes #652

### DIFF
--- a/common/components/constants/LinkConstants.js
+++ b/common/components/constants/LinkConstants.js
@@ -12,7 +12,7 @@ export type LinkSourceDisplayConfig = {|
   +iconClass: string,
 |};
 
-const httpsWwwPrefix = "^https?://w*.?";
+const httpsWwwPrefix = "^https?://\\w*\\.?";
 
 export const LinkDisplayConfigurationByUrl: $ReadOnlyArray<LinkSourceDisplayConfig> = [
   {
--- a/common/components/constants/LinkConstants.js
+++ b/common/components/constants/LinkConstants.js
@@ -12,7 +12,7 @@ export type LinkSourceDisplayConfig = {|
   +iconClass: string,
 |};
 
-const httpWwwPrefix = /^http:s?\/\/w*\.?/;
+const httpsWwwPrefix = "^http:s?\/\/w*\.?";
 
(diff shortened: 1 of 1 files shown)


common/components/constants/LinkConstants.js
```diff
@@ -5,62 +5,63 @@ import type { Dictionary, KeyValuePair } from "../types/Generics.jsx";
...
 
-
 export type LinkSourceDisplayConfig = {|
   +sourceUrlPattern: ?RegExp,
   +sourceDisplayName: ?string,
   +sourceTypeDisplayName: string,
   +iconClass: string,
 |};
 
+const httpsWwwPrefix = "^https?://w*.?";
```
marlonkeating, line 15:
```suggestion
const httpsWwwPrefix = "^https?://\\w*\\.?";
```
In looking at our link urls, I notice that we have some that don't have the www subdomain, like 'drive.google.com' and  'join.slack.com'

marlonkeating (changes requested):

marlonkeating (approved):

Issue #643, opened 2021-06-01
Edit Event page not rendering in Safari


bhg:
When editing an existing event (such as that belonging to Test Admin), a blank page is showing in Safari.

PR #907, opened 2022-09-25
fixed issue #643 - edit event page now renders properly in safari

extreme-coder:
Closes issue #643. Safari doesnt support YYYY-MM-DD date formats.

### DIFF
--- a/civictechprojects/models.py
+++ b/civictechprojects/models.py
@@ -478,9 +478,9 @@ class Event(Archived):
         event = {
             'event_agenda': self.event_agenda,
             'event_creator': self.event_creator.id,
-            'event_date_end': self.event_date_end.__str__(),
-            'event_date_modified': self.event_date_modified.__str__(),
-            'event_date_start': self.event_date_start.__str__(),
+            'event_date_end': self.event_date_end.isoformat(timespec='seconds'),
+            'event_date_modified': self.event_date_modified.isoformat(timespec='seconds'),
+            'event_date_start': self.event_date_start.isoformat(timespec='seconds'),
             'event_description': self.event_description,
             'event_files': list(map(lambda file: file.to_json(), other_files)),
             'event_id': self.id,


PeterBreen (changes requested):
While this does solve Safari refusing to parse `YYYY-MM-DD` strings, it introduces some potential error. 

Use of slashes is non-standard and so we'd be relying on other browsers interpreting a non-standard datetime string correctly; unfortunately, they don't. In my tests with Firefox and Chrome, `Date.parse("2022-09-25")` will return September 25, 2022, midnight UTC. But `Date.parse("2022/09/25")` will return September 25, 2022, midnight __local.__ For me, that's seven hours off. 

see this example from Chrome:
image

Instead, what I'd suggest is finding out if we can continue to use a standards-compliant datetime string, but one that Safari will actually accept along with all other browsers. 

I don't know what the best answer is, but I have three ideas, if they will help.

One possibility is getting further datetime precision from django and passing that in: Safari will parse a datetime string that includes hours, minutes, seconds, and timezone offset. Here's something from Safari for Windows that parsed correctly:

image

Another would be using a library to avoid the Date object; we already load react-moment for other dates and reusing it here might do the job.

A third would be evaluating if [Date.UTC](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/UTC) will help; Safari seems to handle that better than regular Date(). 

It's also quite likely there's a fourth idea I haven't thought of at all, so please don't feel constrained by these three if you have a better idea. :)

Thanks for identifying the issue and the work so far! I think you're well on the way to getting this one solved for good.

extreme-coder:
Hi Peter,

Thanks for providing such a detailed review! I just rechecked and it looks like Django is returning both the date and time, specifically in the format 'YYYY-MM-DD-HH:MM:SS+00:00'. For example, I tested both '2023-03-02 00:30:00+00:00' and '2023/03/02 00:30:00+00:00', and parsing both returns the same result for time. This is true for Chrome and Firefox as well.

PeterBreen:
Ah, thanks for checking what Django is returning. Based on that, I would still prefer pursuing option 1. Dashes are specifically [ISO 8601 compliant](https://tc39.es/ecma262/#sec-date-time-string-format) and slashes are not; so while it's good the UTC vs local issue I found with date.parse() isn't relevant, I think we need to be spec-compliant to solve this for good. Otherwise we're just hoping we get lucky browsers won't ever change how they handle non-standard date strings.

Speaking of getting lucky, though, based on what you found, it seems _none_ of our date strings have been compliant so far. I looked in react devtools and found strings formatted just like you did, `2023-03-02 00:30:00+00:00` -- which isn't right. What we'd need to do is replace the space between DD and HH with the character T, like so: `2023-03-02T00:30:00+00:00` to meet spec. All browsers parse that in my (quick) test.

Please give that a try and let me know if Safari takes it successfully.

extreme-coder:
Turns out the hydrate to JSON function in models.py was using Python's default Date-to-string converter, which doesn't return an ISO-formatted string. I fixed it to use Python's ISO format function, which returns dates with the format you mentioned. I've tested and it runs fine on Safari.

PeterBreen (approved):

PeterBreen:
Tested this in Safari and it works great. Thanks for getting this one solved!

Issue #896, opened 2022-08-09
Perceived failure to save profile picture


doub1ejack:
Found a small UX thing when updating my profile pic. It would not work if I clicked Save before I clicked Done Cropping. I tried & failed to update my photo the other day.

For a quick fix, perhaps the save button could be hidden or disabled until the required cropping has happened. Etc. Thanks!

<img src="https://user-images.githubusercontent.com/1809882/183705446-08dc6b0a-bb86-406d-87b9-4f6f94d77070.png" width="300">

schaferyan:
I'd be interested in looking into this one

marlonkeating:
The library we use is [react-image-crop](https://github.com/DominicTobias/react-image-crop), which we wrap with the component:
https://github.com/DemocracyLab/CivicTechExchange/blob/5a0f28ddb62e7dd57ca3baebfcc43f805c807245/common/components/common/upload/ImageCropUploadButton.jsx

ddfridley:
@schaferyan Hi, I'm helping with development focus for DemocarcyLab. We've created a [DemocracyLab Dashboard](https://github.com/orgs/DemocracyLab/projects/1/views/1?layout=board) and would like to update the status of this issue.

There is an unresoved comment in the PR #911 
> Changing project thumbnails fails with the following console error:
ImageCropUploadButton.jsx:123 Uncaught TypeError: this.props._onIsCroppingChanged is not a function at ImageCropUploadButton._handleFileSelection (ImageCropUploadButton.jsx:123:16) at FileReader.<anonymous> (FileSelectButton.jsx:71:18)

Are you still working on this? If so, what are the next steps, or what challenges are there to moving forward?
Or is this still an issue/should we close it?
Or, if you just aren't available to work on it anymore, please let me know, and/or unassign yourself
Or something else?

Thanks.

schaferyan:
Hi David,

Thanks for reaching out! I''m sorry, I haven't had time to work on
Democracy Lab stuff recently.

I can't recall the details of that issue and I unfortunately don't have an
environment set up to check.

Best,
Ryan

Nostromos:
@ddfridley Happy to take this if its still available, which it seems to be.

ddfridley:
[...]

PR #911, opened 2022-10-01
Crop profile pic #896

schaferyan:
To disable the "Save" button until the user clicks "Done Cropping", I lifted the state variable "isCropping" up from ImageCropUploadButton to EditUserThumbnailModal, and added an _onIsCroppingChanged handler function to EditUserThumbnailModal. This appears to work as expected with no console errors unwanted side effects.

Note that VSCode complains about the type declarations and function binding in many of the .jsx files. I improved this in a few of the files I touched, but addressing this further feels like a separate issue.

### DIFF
--- a/common/components/common/FileInfo.jsx
+++ b/common/components/common/FileInfo.jsx
@@ -1,3 +1,5 @@
+//@flow
+
 export type FileInfo = {|
   id: ?number,
   key: string,
--- a/common/components/common/upload/ImageCropUploadButton.jsx
+++ b/common/components/common/upload/ImageCropUploadButton.jsx
@@ -16,22 +16,24 @@ type Props = {|
   aspect: ?number,
   currentImage: FileInfo,
   isCropping: boolean,
-  onIsCroppingChanged: (boolean) => void,
+  _onIsCroppingChanged: (boolean) => void,
 |};
 
 type State = {|
   s3Key: string,
+  src: string,
   croppedImageUrl: ?string,
+  crop: any
 |};
 
-let lastFileUploadUrl: string = null;
+let lastFileUploadUrl: string = "";
 
 class ImageCropUploadButton extends React.PureComponent<Props, State> {
   constructor(props): void {
     super(props);
     this.state = {
       s3Key: "",
-      src: null,
+      src: "",
       croppedImageUrl: "",
       crop: this.initializeCropSettings(),
     };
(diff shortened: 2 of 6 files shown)


PeterBreen (commented):
Looks pretty good to me, and I've done a test of cropping where the save button is disabled while the crop window is open. 

I did run into an issue I'd like to know if you can replicate. Here's the steps:
 - log in, go to user profile, edit profile image
 - click upload new image, select any image but do not crop
 - click the modal's Cancel button (not "cancel" for the image crop, but the "close this entire modal" cancel at the bottom)
 - click the edit image icon again to re-open this cropping modal
 
What I get is this:
image

this persists until I refresh the page. I can't replicate this on prod-mirror or staging but I can replicate it in my dev environment. Any idea what's going on there?

schaferyan:
Thanks, when I follow those steps this is happening in my developer environment as well. I'm not sure of the cause but I will look into it!

schaferyan:
Haven't had a chance to verify this, but I suspect it is related to the state variable `croppedImageUrl` in `ImageCropUploadButton`. Cropping, canceling cropping, or saving crop all update `croppedImageUrl`. The default value of the variable is an empty string, and I imagine that if the modal's cancel button is clicked the empty string is stored as the location of the thumbnail - casing the image not to be found when the modal is opened again.

schaferyan:
@PeterBreen I believe I have solved this. The value of the state variable `isCropping` in `EditUserThumbnailModal` needed to be set to false when the modal is closed. Not sure why this wouldn't show up on prod-mirror or staging. In any case, I'm not seeing the problem in my dev environment anymore, let me if this is true for you as well!

PeterBreen:
Sorry, I think I was unclear in my reference to prod-mirror and staging; I meant to say (but failed entirely) that it seemed to be an issue with this PR vs current main branch state. 

Either way I pulled the latest version of this branch, ran my list of steps and can't reproduce the error, so I think you got it fixed too. Nice, thank you!

PeterBreen (approved):
The only issue I had was resolved; looks good!

marlonkeating (changes requested):
Changing project thumbnails fails with the following console error:
`ImageCropUploadButton.jsx:123 Uncaught TypeError: this.props._onIsCroppingChanged is not a function
    at ImageCropUploadButton._handleFileSelection (ImageCropUploadButton.jsx:123:16)
    at FileReader.<anonymous> (FileSelectButton.jsx:71:18)`

PR #913, opened 2022-10-04
Pre-commit: Add prettier support

naqeeb:
### Description
Pre-commit will now also run prettier on CSS and JS files.

### Implementation Notes
1. prettier will only run on CSS and JS files. Let me know if you'd like to include JSX files as well
2. This will use the prettier configs specified in  `.prettierrc.json`

### Additional Information

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,3 +10,9 @@ repos:
     rev: 22.8.0
     hooks:
     -   id: black
+-   repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v2.7.1 # Use the sha or tag you want to point at
+    hooks:
+      - id: prettier
+        types_or: [css, javascript]
+        args: [--config=.prettierrc.json]


PeterBreen (dismissed):
Please do add .jsx; we use that extension a lot, so it's important to have. If you haven't already seen it, the discussion in #532 may have some relevant context for prettier config.

PeterBreen:
What a commit name - I meant "instead of 3.10.11" which errored, "3.10" does not. Trying to match our heroku python version -- but I'm not sure we need to specify default_language_version ?

That aside,  there are some additional steps required to get this to actually _do_ anything, when testing this branch (or using it)
```
pip install pre-commit
pre-commit --version  // should return 3.4.0 or higher
pre-commit install
```

only then will it start to change files on commit. We should probably roll that into initial setup or something so every dev doesn't have to do this manually.

For testing purposes, what we should see is: scss, css, jsx, js files are formatted by prettier. All other file types, such as py, htm, html, should _not_ be changed by prettier.

Issue #468, opened 2020-11-30
Find Projects: Clicking filter makes other selected filters flash white for a second while new search results load


glenbraun:
Repro Steps:
- Navigate to Find Proejcts page
- Select a filter
- Select another filter

Expected: The first selected filter stays visibly selected after clicking the second filter
Actual: The first selected filter looks like it is unselected for a moment while the new search results load

[https://trello.com/c/wQ7mhDRU](url)

**Peter Breen Oct 15, 2019 at 10:10 AM (edited)**
Based on several attempts at solving this, I think the problem lies in the filter checkbox state management being a "am I in this selectedTags object" function instead of a direct true/false key which we then manipulate with setState.

Or put another way, React is re-rendering every single filter on every single change of any filter because every time a change occurs every tag has to ask 'am I selected?' to know whether or not to read checked or not. It has to; otherwise a given filter has no idea if it should uncheck itself (say a user clicked Clear Filters) or check itself (a user clicked a filter in the list)

To solve this, I think we will need to refactor the flux store for the filters, without breaking the Clear Filters button and the tag pills. Rather than holding a selectedTags object, it should output a list of true/false key value pairs for every single filter, and then calculateState should set true or false as required for individual keys on change. One of my tests had the filter components manage state this way and the blinking/flash was completely gone, though unfortunately it did so by ignoring flux as the 'source of truth.' Combining flux as the source of truth with that setup should work.

for example rather than a filter having checked={this.props.checkEnabled(tag)} it would be something like checked={this.state.nameOfTag} and simply updating state would update the filter. Or more critically, not updating state would not update and rerender a given filter.

I'm not sure if it would be best to hold all of these keys in the parent (ProjectFilterDataContainer.jsx) component or in individual child (RenderFilterCategory.jsx) components.

PR #918, opened 2022-10-08
fixed issue #468; find projects filters now work properly

extreme-coder:
Resolves issue #468. Load projects was clearing and refilling all project data, including filters, which don't need to be refreshed.


PeterBreen (approved):
This is currently available for testing on prod-mirror, and the difference is (for me) _very_ noticeable, especially when selecting filters with large numbers of projects and/or multiple filters at once. Great work, thank you!

I've done a quick pass on making sure we don't have unintended issues on other pages which display projects (groups, events, homepage) and so far so good there as well.

PR #920, opened 2022-10-25
Added coverage for js on host

lmpeixoto:
To calculate the corverage run the script coverage-js.sh on unix environment, or alternatively this command:
`npm run test-cover`

to do: add documentation

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@ Pipfile
 Pipfile.lock
 
 # Unit test / coverage reports
+coverage/
 htmlcov/
 .tox/
 .coverage
--- a/coverage-js.sh
+++ b/coverage-js.sh
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+npm run test-cover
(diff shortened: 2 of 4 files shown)


PeterBreen (commented):
Hi, sorry about the delay here - reviewing this is really a task for Marlon and he's busy, but I took a look at this to try and help out. I've got a couple questions:

 - Maybe I'm missing it, but I don't see an overall coverage number.  My output says 56.02% lines are covered and I _know_ that's not right. 56% of what we have tests for, maybe, but not our entire codebase. Is that something that can be added?
 
 - running tests creates a `coverage/` directory; I think that should be added to .gitignore unless we need it in the repo?

lmpeixoto:
Hi Peter thanks for reviewing.

As far as the report is concerned I get the following output:

image

My reading from that is that the JS code has 86.97% statement coverage and 87.5% branch coverage. Tbh my branch is a little bit outdated.

You're right I'll include the .gitignore for coverage folder.

PeterBreen:
Closing due to age; issue still open (and this is still useful reference)

PR #929, opened 2022-11-21
Filter out private projects when computing tag counts - issue#924

extreme-coder:


PeterBreen:
Could you tell me the steps you took to test this? I'm trying to test myself and it's not working.

 Here's what I'm doing:
git checkout issue_924
Login to django admin control panel and set a project to private. 
Check django console to see if the cache regenerates. 
npm run dev (just in case)
Browse to Find Projects, and see if the filter count updates.

That last step is where it doesn't work. For example I set a project with the "Political Reform" tag to private. My dev environment has 4 projects with that tag. The filters still show a count of 4, but when the filter is selected only 3 return - so the tag count isn't updating properly for me.

I looked at the raw API return and I'm wondering if this is the problem:
image
 
our python structure is not my area of expertise, but I think that's from here: https://github.com/DemocracyLab/CivicTechExchange/blob/master/civictechprojects/helpers/search/projects.py#L196

anyway please let me know if I'm doing something incorrectly here.

extreme-coder:
I have found the problem, it seems tags cache was not getting updated when a project is updated.
There were two differences in how I tested. 1) I didn't use the admin panel to update the project, I used the Edit project feature to update. 2) I think I restarted the django server in between the project update which cleared the tags cache.

I have updated my commit to include a fix to update the tags cache when a project is updated. 

I have now tested by both updating the project from admin panel and Edit project feature.

PeterBreen (approved):
It works for me now, thank you. After setting projects to private and refreshing the find projects page, tag counts update as expected.

Issue #925, opened 2022-11-16
Filters wrongly shown for project's hidden roles


marlonkeating:
Currently, projects' hidden roles are showing up in filters (even though filtering by these hidden roles will fail to find the projects in question).  We need to filter out hidden roles when generating the project search page role filters.

## Code Location
https://github.com/DemocracyLab/CivicTechExchange/blob/e87788d5d3df1226d61877a621d95e6ecc5aaf55/civictechprojects/caching/cache.py#L137

PR #930, opened 2022-11-21
exclude hidden roles from project filter - issue #925

extreme-coder:


civictechprojects/caching/cache.py
```diff
@@ -134,6 +134,8 @@ def _projects_tag_counts(event=None, group=None):
                 organization_type += project.project_organization_type.slugs()
 
                 project_positions = project.get_project_positions()
+                # exclude roles which are hidden
+                project_positions = filter(lambda position: not position.is_hidden, project_positions)
```
marlonkeating, line 138: This technically works, but our preferred way of filtering is to leverage django's ORM.  

```suggestion
                project_positions = project.get_project_positions().filter(is_hidden=False)            
```

marlonkeating (changes requested):

civictechprojects/caching/cache.py
extreme-coder: Updated the code and tested

marlonkeating (approved):

Issue #636, opened 2021-05-22
Cannot delete a second project/group due to broken confirmation modal


marlonkeating:
On the My Projects page, if a user deletes a project and then tries to delete a second project, or they cancel out of the first delete operation, they will be greeted by a broken delete modal upon their next attempt.

![image](https://user-images.githubusercontent.com/322346/119242852-4e2a4600-bb16-11eb-8f31-e501afddae84.png)

schaferyan:
I encountered this while working on #514. Marking the `confirmDeleteProject` method in `MyProjectsController` as `async` seems to resolve the issue. If you want to add me to this issue I can submit a pull request for it, otherwise the fix can be included when I submit a PR for #514 (which will likely take some time) if that works.

PeterBreen:
If you've got a solution for this, then by all means please submit a PR. I assigned you to the task just now, as you probably saw.

PR #931, opened 2022-11-27
Fixed: Cannot delete a second project/group due to broken confirmation modal

schaferyan:
Adding `async` to the method signature of `comfirmDeleteProject()`  in both `MyProjectController` and `MyGroupsController` allows `ConfirmationModal` to receive the Promise object it expects, clearing a console error and resolving Issue #636


PeterBreen (approved):
I tested this with #636's "try to delete, cancel, try to delete" method. I first made sure the problem happened on main (still does) and then replicated the exact same steps in this branch, and had no issue opening and cancelling as many confirmDelete modals as I wanted.

PR #935, opened 2022-12-02
Home: Show full date on next hackathon card

### LINKED ISSUE #934: Home: Upcoming event card should show full date
The upcoming event card on the home page shows only the time of the upcoming event, not the date.  This was by design for the Event listings page, which lists the date headers above the cards, but these headers don't exist on the home page.

marlonkeating:
resolves #934

### DIFF
--- a/common/components/componentsBySection/FindEvents/EventCard.jsx
+++ b/common/components/componentsBySection/FindEvents/EventCard.jsx
@@ -31,7 +31,7 @@ class EventCard extends React.PureComponent<Props> {
         >
           {this._renderEventLogo()}
           {this._renderEventInformation()}
-          {this._renderEventTime()}
+          {this.props.showFullDate ? this._renderStartEndDateTime() : this._renderEventTime()}
         </a>
       </div>
     );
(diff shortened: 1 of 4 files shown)


PeterBreen (dismissed):
ed: whoops, reviewed too early

common/components/componentsBySection/FindEvents/EventCard.jsx
```diff
@@ -71,12 +72,33 @@ class EventCard extends React.PureComponent<Props> {
...
           <h2>
-            <Moment format="LT">{Event.event_date_start}</Moment>
+            <p>
+              Begins:<br/>
+              <Moment format={dateTimeFormat}>{event.event_date_start}</Moment>
+            </p>
+            <p>
+              Ends:<br/>
+              <Moment format={dateTimeFormat}>{event.event_date_end}</Moment>
+            </p>
```
PeterBreen, line 101: If we remove the containing H2, we can avoid <br> and also style the headers and dates differently. 

I think giving "Begins:" and "Ends:" the same font size/weight (\<h2>) as the event title makes the card visually lopsided with all the added date information there, so I'd suggest this:

```
<h3>Begins:</h3>
<p><Moment format .... </Moment></p>
<h3>Ends:</h3>
<p><Moment format .... </Moment></p>
```

and in order to fix some spacing issues that come up here,
https://github.com/DemocracyLab/CivicTechExchange/blob/master/civictechprojects/static/css/partials/_FindEventController.scss#L31 change this to `.EventCard-root h2, .EventCard-root h3`


And then to avoid an issue with long location strings, like this:
image

https://github.com/DemocracyLab/CivicTechExchange/blob/master/civictechprojects/static/css/partials/_FindEventController.scss#L109-L112
add `margin-right: 0.5rem;` here



the end result should be this:
image

PeterBreen (changes requested):
one requested change for style, which was - as CSS tends to be - "change this one thing. Oh wait that needs another fix. Oh wait that needs another, another fix."

PeterBreen (approved):

PR #937, opened 2022-12-07
chore: enable loading dotenv in manage.py

crazyoptimist:
added django-dotenv package


crazyoptimist:
dot env file is only for containers as of now. use shell script for exporting env vars.

PR #943, opened 2023-01-26
Support for changing Event zoom room links + Event Fixes

marlonkeating:
Assorted fixes to support Events

## Caching 
- Event Rooms can be altered on the admin page, and they now refresh the event cache so changes are immediately viewable

### Home Page 
- Upcoming event is now shown until the end of the event, not the beginning of the event
### Event Page
- RSVP Modal: Sign up/Cancel buttons now shown alongside Join Video button
- RSVP Modal: Join Video button no longer shown when event has been deactivated


PeterBreen (approved):
Got that last bit of feedback handled (thanks, Marlon) and the rest still looks just fine as it always did.

PR #944, opened 2023-02-05
Change font load from @import to <link>

PeterBreen:
note: blocked by #960. Will reopen and finish after that's merged.

 - Change how we load google fonts for improved performance (currently loading our site js is required to begin loading google fonts, this fixes that)
 - adds preconnect to FontAwesome to help speed up loading of FA glyphs

Why link instead of import? See this SO thread: https://stackoverflow.com/questions/12316501/including-google-fonts-link-or-import but the short version is, this lets us load in parallel instead of in sequence.

Lighthouse makes the chaining issue clearer visually:
image

Another option I found is self-hosting the font files, which is even suggested by google in some cases: https://www.youtube.com/watch?v=Mv-l3-tJgGk&t=1498s -- but that'll be for another PR.

### Results:

I plan on recording page loads on prod-mirror as the control, and on staging with this PR deployed as the test. Caching disabled for all tests.


PeterBreen:
Some of the commits here are me trying to troubleshoot heroku's CLI offering this error:
build_error_console


https://devcenter.heroku.com/articles/python-support#checking-the-python-buildpack-version
https://devcenter.heroku.com/articles/python-support

so this PR is messy and I'll clean it up before asking for a review, but heroku's admin panel and CLI are both rejecting builds right now and I don't like it!

PeterBreen:
Deleting this and redoing the PR to not have a messy commit history. My fault for committing a bunch of nonsense to this pr testing heroku's deployment issues.

PR #948, opened 2023-02-28
Salesforce integration

justBanks:


PeterBreen:
@justBanks there's this salesforce PR and also #891. Is it safe to close that one and keep this open, or do we need both? 

i'm trying to do some cleanup on our open PRs/issues, is why I'm asking

PR #951, opened 2023-03-19
ad showHeader props to hide footer

andrewindechemain:


example.env
marlonkeating, line 1: Why is this file deleted?

common/components/componentsBySection/CreateEvent/AboutEventDisplay.jsx
```diff
@@ -37,6 +37,7 @@ import EventProjectRSVPModal from "../../common/event_projects/EventProjectRSVPM
 type Props = {|
   event: ?EventData,
   viewOnly: boolean,
+  showHeaders: false
```
marlonkeating, line 40: This should be a boolean value unless we know for a fact it will always be false.
```suggestion
  showHeaders: boolean
```

common/components/componentsBySection/CreateEvent/AboutEventDisplay.jsx
```diff
@@ -85,6 +86,7 @@ class AboutEventDisplay extends React.PureComponent<Props, State> {
       showPostRSVPModal: false,
       showCancelRSVPConfirmModal: false,
       startDate: startDate,
+      showHeaders: false,
```
marlonkeating, line 89: Where is this state value being used in the code?

marlonkeating (changes requested):

example.env
andrewindechemain: Hello,I have returned it back.I had used it to create an env.

common/components/componentsBySection/CreateEvent/AboutEventDisplay.jsx
andrewindechemain: Okay,I have changed the showHeaders to a boolean value

common/components/componentsBySection/CreateEvent/AboutEventDisplay.jsx
andrewindechemain: I have chaned the original state of showHeaders to have a boolean value. If I am not wrong this value will change to false on this key.

PR #960, opened 2023-04-08
Library Updates to support Heroku-22

marlonkeating:
This change includes the library upgrades needed to upgrade our heroku stack from Heroku-18 to Heroku-22

## Upgrades
- Python 3.6.7 -> 3.10.11
- Node 14.15 -> 16.20

## Python Deprecations
- django.conf.urls.url -> django.urls.re_path
- requests.is_ajax -> common.helpers.request_helpers.is_ajax


PeterBreen (approved):
The VM version of this works just fine; booted it up and found a couple minor issues that just need an extra step on first use (and only first use) -- neither of which are blocking. Otherwise, it brings up dlab and deployment, as shown in this branch history, works. 

I haven't tried Docker - but I haven't seen any issues from other devs with Docker. Making sure that works would be nice, but it works for me as-is. 👍

Issue #957, opened 2023-04-03
Disable carousel autoadvance


PeterBreen:
Per the accessibility team, we should disable the carousel automatic advance, meaning the only way to advance a slide is manually.

Our version of react-bootstrap (1.6, using Bootstrap 4.x) may not support `interval={null}` to disable autoadvance; that seems to be a react-bootstrap 2.x/Bootstrap 5.x function. However, if we can use null, let's do so.

To close this item, two things are required:
- remove the ability to set a custom interval via props for the carousel components
- remove any props.intervals from instances of those components.
- test that carousels effectively do not auto advance, and only change slides on user action (clicking the arrow)

If interval={null} does not work, set interval={315569520000}. When I discussed this with the accessibility team, they allowed as how users are _probably_ not going to wait ten years for a slide to advance. ;)

We have two Carousel components which need work: TestimonialCarousel and BlogCarousel.

## code locations
First, remove the logic allowing for custom intervals and simply have it set it to `interval={null}`, then remove interval from our type statement; we won't accept custom values via props anymore.

TestimonialCarousel:

 - https://github.com/DemocracyLab/CivicTechExchange/blob/1deba2dff9a48802e763b7e415af571207dfc37b/common/components/common/carousel/TestimonialCarousel.jsx#L37
 - https://github.com/DemocracyLab/CivicTechExchange/blob/1deba2dff9a48802e763b7e415af571207dfc37b/common/components/common/carousel/TestimonialCarousel.jsx#L8-L9

BlogCarousel:
 - https://github.com/DemocracyLab/CivicTechExchange/blob/1deba2dff9a48802e763b7e415af571207dfc37b/common/components/common/carousel/BlogCarousel.jsx#L36
 -  https://github.com/DemocracyLab/CivicTechExchange/blob/1deba2dff9a48802e763b7e415af571207dfc37b/common/components/common/carousel/BlogCarousel.jsx#L8-L9

Second, remove any use of props.interval:

  - https://github.com/DemocracyLab/CivicTechExchange/blob/2c5651d6b3a85f750e06a903a0c47d80a3c9ebdf/common/components/controllers/CorporateHackathonController.jsx#L264-L266
  - https://github.com/DemocracyLab/CivicTechExchange/blob/2c5651d6b3a85f750e06a903a0c47d80a3c9ebdf/common/components/controllers/CorporateHackathonController.jsx#L410-L413
[...]

PR #962, opened 2023-04-11
Xuanying #957 disable carousel auto advance

SleeplessRegina:
Disable carousel auto advance, make the carousel only support manually advance both in TestimonialCarousel and BlogCarousel.
Set the custom intervals to null and remove interval from our type statement.


common/components/controllers/CorporateHackathonController.jsx
```diff
@@ -263,7 +263,7 @@ class CorporateHackathonController extends React.PureComponent<{||}, State> {
           <div className="corporate-hackathon-carousel carousel-testimonial-root">
             <TestimonialCarousel
               category="partner-highlights"
-              interval={600000}
+              interval={null}
```
PeterBreen, line 266: Do we need to use interval={null} here, as well as in the TestimonialCarousel component itself? Carousel shouldn't be taking any props so I'm hoping it's safe to remove, but please test and let me know?

common/components/controllers/CorporateHackathonController.jsx
```diff
@@ -409,7 +409,7 @@ class CorporateHackathonController extends React.PureComponent<{||}, State> {
           <div className="corporate-hackathon-carousel carousel-testimonial-root">
             <TestimonialCarousel
               category="hackathon-highlights"
-              interval={600000}
+              interval={null}
```
PeterBreen, line 412: Same question as above

common/components/controllers/LandingController.jsx
```diff
@@ -363,7 +363,7 @@ class LandingController extends React.PureComponent<{||}, State> {
         <h2 className="text-center LandingController-section-header">
           What people are saying about DemocracyLab
         </h2>
-        <TestimonialCarousel interval={15000} />
+        <TestimonialCarousel interval={null} />
```
PeterBreen, line 366: Same question as above

PeterBreen (changes requested):
It looks good, but please see the three line comments on instances of using interval as a prop that we may not need. But please do let me know if it is necessary :)

SleeplessRegina:
> It looks good, but please see the three line comments on instances of using interval as a prop that we may not need. But please do let me know if it is necessary :)

Yes, I removed the interval props of TestimonialCarousel then tested again and found it still looks good. Both the carousel won't automatically advance.

PeterBreen (approved):
Thanks for those changes; looks good, works exactly as expected 👍

PR #969, opened 2023-05-05
589 log on a single line

Workhqd:
When export DJANGO_DEBUG='False' or export DJANGO_DEBUG=False, error message could be printed in one line.
image

In other situation, error message will be printed in multi line.
image


marlonkeating (changes requested):
Looking at the logging results, I think it would be helpful to still see newline indications, so instead of replacing newline characters with nothing, perhaps a literal '\n' printed out would be good?  I'm open to other suggestions.

PR #970, opened 2023-05-06
remove looks good

Workhqd:
Now it would not show "looks good"

6229a8670577c776690742bb7abc92c

(However, it will still shows ✓. should I remove ✓? Or is it enough now?)

By the way, please run "npm run build" to apply the change. The Node.js version should be 16.20.x. If the node.js 's version is other version, such as 18.x.x(the default version in ubuntu 22), please change the version to 16.20.x, then you could run "npm run build" successfully.

how to use node.js 16.20.x:
curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.1/install.sh | bash
nvm install 16.20
nvm use 16.20

### DIFF
--- a/common/components/controllers/LogInController.jsx
+++ b/common/components/controllers/LogInController.jsx
@@ -78,7 +78,7 @@ class LogInController extends React.Component<Props, State> {
               placeholder="Email"
               onChange={e => this.setState({ username: e.target.value })}
             />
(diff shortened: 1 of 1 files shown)


PeterBreen (approved):
Looks good, works in my tests. I saw the question about the green check -- I think we can keep that for now.

Workhqd:
Thank you very much!

PR #981, opened 2023-06-22
Docker gdal

ddfridley:
Latest version of master needs GDAL for python.  This change to Dockerfile install's it.


ddfridley:
I have tested this in a PC environment. It would be good to have someone else test it, and in some other environments.

ddfridley:
Don't merge this PR - because of the bashrc-setup file.  I'll create another one.

PR #982, opened 2023-06-23
what it takes to install gdal for python in docker

ddfridley:
Here is what I put in to Dockerfile to get gdal working. 
Here is a link to the getting started instructions for windows that I wrote https://docs.google.com/document/d/1VzMtjhqUh5QEuxXq_m8mDXHHPaVySnnfiqLA15ebov8/edit?usp=sharing

PR #984, opened 2023-06-26
Iframes#978

ddfridley:
Also - one issue.  The path is /igs instead of /igroups as was in the spec.  The problem is that I originally tried to do this as a sitemap - using igroups modeled after the groups path.  But I learned it's better to do it as just a path. BUT even after deleting all the code around the /igroups path - it still routs to that path. It seems to be cached somethere that I haven't figures out yet.  So I can't change the path to /igroups and get it to run my code.  Any ideas? 

If you build this code and run it locally, then you can brows to the file `./iframe-test.html` and it will load the projects as an iframe:

image

There might also be a better place to put that test file than root - I'm open to where.


ddfridley:
Here's what it looks like embedded in the Open San Diego jekyll page: 
image

And here's the code that it took to embed it there:
```
        <div style="position: relative; overflow: hidden; width: auto; margin-left: -3rem; margin-right: -3rem; height: 83rem; border-radius:1rem; border: 1px solid black;">
            <iframe
                src="http://localhost:8000/igs/1"
                style="position: absolute; top: 0; left: 0; bottom: 0; right: 0; width: 100%; height: 100%; border: none; border-radius: 1rem;"
            ></iframe>
        </div>
```

ddfridley:
iframe resizer has been added and the ips path for rendering projects in an iframe is implemented.  This is ready for merging.
But there are two things that have been changed that need to be considered:

- [ ] The [Dockerfile](https://github.com/DemocracyLab/CivicTechExchange/pull/984/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557) has the changes from #982 that are required for me to run this on docker on my machine.  I agree, it should not be different on different machines.  On the other hand, other's have had trouble with requirements.py #989.  I think there is some python or linux version discrepancy.   One option is to take out this change from the PR and see if it can work in qa and production - there is already a separate PR for it.

- [ ] In [settings.py](https://github.com/DemocracyLab/CivicTechExchange/pull/984/files#diff-713423226d71e344a140450b374773dfa07d2de05ecb54db9f1c588967067767) I have made the  CACHES conditional on the DEBUG environment.  Not caching helped with a problem where web pages don't change after I change the code.  But perhaps there is another way to do this, or perhaps in QA you need CACEHES enabled and DEBUG.  

- [ ] There are 2 test*.html files in the root directory that maybe should go somewhere else.

+ I would like to get this into QA at some public URL so I can test for cross-site permissions problems.

ddfridley:
I pulled the latest master, and added opensandiego.org to CSP_FRAME_SRC and CSP_FRAME_ANCESTORS in the democracy-lab-dev heroku settings and then deployed it there. It works, and I tested it with the opensandiego.org page to make sure it works.  -- but now I have updated the opensandiego page to point to production so it won't show the incorrect info, 

A better resolution to manually setting the CSP_... env variables would be to ask leader to enter the URL of their web site when they create their groups, and projects, and then automatically add those domains to the CSP variables - but for now the number of sites will be small, and we should get this feature out their and get feedback before making it easier to administer.

So I think this is ready to go.

common/components/common/groups/IframeGroupDisplay.jsx
```diff
@@ -0,0 +1,35 @@
+// @flow
+// derived from AboutGroupDisplay
+import React from "react";
+import { Container } from "flux/utils";
+import AboutGroupDisplay from "./AboutGroupDisplay.jsx"
+import type { GroupDetailsAPIData } from "../../utils/GroupAPIUtils.js";
+import ProjectCardsContainer from "../../componentsBySection/FindProjects/ProjectCardsContainer.jsx";
+
+class IframeGroupDisplay extends AboutGroupDisplay<Props, State> {
```
marlonkeating, line 9: Why does this need to inherit from AboutGroupDisplay?

common/components/common/projects/AboutProjectDisplay.jsx
```diff
@@ -108,8 +108,9 @@ class AboutProjectDisplay extends React.PureComponent<Props, State> {
 
   render(): React$Node {
     const project = this.state.project;
+    const widthModifier=window.parent !== window ? ' use-parent-width' : '';
```
marlonkeating, line 111: The `window.parent !== window` check is made in a lot of places, it should live in a helper function.

common/components/componentsBySection/FindProjects/ProjectCard.jsx
```diff
@@ -39,6 +39,10 @@ class ProjectCard extends React.PureComponent<Props, State> {
       }),
       showModal: false,
     };
+    // iFrameResizer will startup after the page is rendered, so we need to rerener if it does
+    if(!window.iFrameResizer) window.iFrameResizer={}
+    if(!window.iFrameResizer.onInParent) window.iFrameResizer.onInParent=[]
+    window.iFrameResizer.onInParent.push(this.onIframeResizer.bind(this))
```
marlonkeating, line 45: This should live in a helper function or wrapper component. 
Also, don't forget line semicolons!

common/components/controllers/MainController.jsx
```diff
@@ -24,6 +24,35 @@ class MainController extends React.Component<{||}, State> {
...
+    // if we are running within an iframe, and the parent is running IframeResizer then do this
+      if(!args[0].data.includes('[iFrameSizer]')) return
+      // let the height go auto rather than being the height defined by the parent - since we will be resizing
+      const style = document.createElement("style")
+      style.textContent = ".Profile-primary-container.frame-full .AboutGroup-card-container .row .ProjectCardContainer .row {height: auto!important;}"
+      document.head.appendChild(style)
+      if(!window.iFrameResizer) window.iFrameResizer={}
+      window.iFrameResizer.inParent=true
+      let func
+      while((func=window.iFrameResizer.onInParent?.shift()))
```
marlonkeating, line 37: What is this line doing?

common/components/urls/urls_v2.json
```diff
@@ -6,6 +6,7 @@
   {"name": "CreateEventProject", "pattern": "events/(?P<event_id>[\\w\\-]+)/projects/create/(?P<project_id>[\\w\\-]*)"},
   {"name": "AboutEventProject", "pattern": "events/(?P<event_id>[\\w\\-]+)/projects/(?P<project_id>[\\w\\-]+)"},
   {"name": "CreateProject", "pattern": "projects/create/(?P<id>[\\w\\-]*)"},
+  {"name": "IframeProject", "pattern": "ips/(?P<id>[\\w\\-]+)"},
```
marlonkeating, line 9: My preference would be for urls of iframe versions of pages to branch off from the main urls.
```suggestion
  {"name": "IframeProject", "pattern": "projects/ips/(?P<id>[\\w\\-]+)"}, 
```

iframe-test-resizer.html
marlonkeating, line 1: Is this a demonstration file for clients wanting to embed projects/groups?  If so, add comments.
Also, these html files should probably live in a separate folder.

marlonkeating (changes requested):

common/components/controllers/AboutGroupController.jsx
```diff
@@ -60,7 +60,7 @@ class AboutGroupController extends React.PureComponent<{||}, State> {
 
   _renderLoading(): React$Node {
     return this.state.statusCode
-      ? this_.renderLoadErrorMessage()
+      ? this._renderLoadErrorMessage()
```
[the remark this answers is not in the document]
ddfridley: Yes. It's a bug fix. `this_` is undefined and causes an exception when executed. Where as `this` is what was intended. I encountered this when testing.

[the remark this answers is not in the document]
marlonkeating: Oh, keen eye!  I looked over this I don't know how many times and didn't notice...

iframe-test-resizer.html
ddfridley: These two files are intended for use be developers in unit testing. I couldn't figure out a directory for putting them in. But they should be retained for future testing of this feature. And I will put comments in them explaining that's what the are for. 

But how would you feel about a `tools` directory at the top level and I would put them in there.  It's doesn't feel right to call it a test directory because that kind of implies automated tests.  Just leaving them at the root isn't very clean. Or I am open to other ideas.

common/components/controllers/MainController.jsx
ddfridley: I will add a comment to the code.  There could be multiple components needing to be re-rendered after iframe resizer starts, so onInParent is an array of callbacks from those components to run after iframeresizer starts.  Components may need to rerender after iframeresizer starts because their style needs to change if rendering in an iframe that resizes v. rendering in an iframe that doesn't resize.

Dockerfile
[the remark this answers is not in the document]
ddfridley: Yes. It doesn't work for me on windows without this.  I think I have gotten a few other people have tried #982 and at least not complained, and had reports of success with the getting started doc too.

common/components/common/groups/IframeGroupDisplay.jsx
ddfridley: This function needs to have its own unique name, and would have a lot of code duplicated from AboutGroupDisplay, but the render function is different.  Doing it this way eliminates duplicate code.

common/components/componentsBySection/FindProjects/ProjectCardsContainer.jsx
```diff
@@ -71,7 +71,7 @@ class ProjectCardsContainer extends React.Component<Props, State> {
           </React.Fragment>
         ) : null}
         <div className="row">
-          {!_.isEmpty(this.state.projects) && (
+          {!_.isEmpty(this.state.projects) && !this.props.supressHeader && (
```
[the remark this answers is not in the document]
ddfridley: done

common/components/common/projects/AboutProjectDisplay.jsx
ddfridley: isWithinIframe - done.

common/components/componentsBySection/FindProjects/ProjectCard.jsx
ddfridley: This was an interesting challenge. I created common/components/common/IframeResizerInParent.jsx a wrapper component with static helper functions that cleanly factored this out.

common/components/urls/urls_v2.json
ddfridley: done but its projects/inframe/ and groups/inframe/

iframe-test-resizer.html
ddfridley: I created a tools directory and put them there.

ddfridley:
Okay I've made and pushed the changes and this is ready to go.

common/components/common/projects/IframeProjectDisplay.jsx
marlonkeating, line 1: Do we need this file?  It seems to be just a wrapper around AboutProjectDisplay (which handles the iframe logic internally).

marlonkeating (changes requested):

common/components/controllers/MainController.jsx
```diff
@@ -78,19 +56,21 @@ class MainController extends React.Component<{||}, State> {
...
-        onMainHeaderHeightChange={this._mainHeaderHeightChange.bind(this)}
-      />}
-      <SectionController
-        key="section_controller"
-        headerHeight={this.state.headerHeight}
-      />
-      {ShowHeadAndFoot && <SponsorFooter key="sponsor_footer" />}
-      {ShowHeadAndFoot && <SiteFooter key="site_footer" />}
-    </>
+    const ShowHeadAndFoot=!(window.location.pathname.includes('/groups/inframe')||window.location.pathname.includes('/projects/inframe'));
```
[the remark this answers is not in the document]
ddfridley: I made the change to urlHelper both in this file and in ProjectCard. - but the code is different between the two and a common helper method wasn't possible.  Plus both places read well with the urlHelper change.   The ProjectCard code looks like this now:
```JS
    const url: string = (urlHelper.atSection(Section.IframeGroup) && IframeResizerInParent.inParent()) ? 
        urlHelper.section(Section.IframeProject, {
          id: this.props.project.slug || this.props.project.id,
        }) :
        ( this.props.project.cardUrl ||
          urlHelper.section(Section.AboutProject, {
            id: this.props.project.slug || this.props.project.id,
        })
      );
```

common/components/utils/isWithinIframe.js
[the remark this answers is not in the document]
ddfridley: done

common/components/common/projects/IframeProjectDisplay.jsx
ddfridley: removed it and IframeProjectDisplay by the same logic.

ddfridley:
I've tested this locally, and pushed it to democracy-lab-dev and am testing it a little more there, but won't finish until tomorrow.

ddfridley:
ok, I'm done with my testing.  ready to go.

ddfridley:
Also, please add opensandiego.org to CSP_FRAME_SRC and CSP_FRAME_ANCESTORS in the environment settings after you push to production.

ddfridley:
More specifically in democracy-lab-dev I have:
```
CSP_FRAME_ANCESTORS=["qiqochat.com", "*.qiqochat.com", "*.google.com", "democracylab.org", "democracy-lab-prod-mirror.herokuapp.com", "democracy-lab-dev.herokuapp.com", "democracy-lab-staging.herokuapp.com", "*.doubleclick.net","https://opensandiego.org/"]
CSP_FRAME_SRC=["qiqochat.com", "*.qiqochat.com", "*.google.com", "*.youtube.com", "democracylab.org", "democracy-lab-prod-mirror.herokuapp.com", "democracy-lab-dev.herokuapp.com", "democracy-lab-staging.herokuapp.com", "*.doubleclick.net", "*.hotjar.com","*.opensandiego.org"]
```
I remember it didn't work without the https:// in front for CSP_FRAME_ANCESTORS

civictechprojects/static/css/partials/_Profile.scss
```diff
@@ -114,6 +114,34 @@
...
+  height: 100vh; // force the flex box to use the whole iframe height that it is in
+}
+
+.Profile-primary-container.frame-full .AboutGroup-card-container .row .ProjectCardContainer .row .col-sm-12.col-lg-6 .ProjectCard-root {
+  margin: 0;
+}
+
+// for the project cards burried in the iframe, give them a fixed width and let them fill the frame
+.Profile-primary-container.frame-full .AboutGroup-card-container .row .ProjectCardContainer .row .col-sm-12.col-lg-6 {
+  flex: 0 0 26rem;
```
PeterBreen, line 138: Declaring a fixed width here is causing some issues, most obviously on any device where the iframe is below 26rem width - like, say, a mobile device. It also makes cards look unusual at very high res, so a more nuanced solution is needed here - perhaps with max-width, media queries? If it's not working on mobile and it looks ... real odd on high res, let's give it another pass?

civictechprojects/static/css/partials/_Profile.scss
```diff
@@ -114,6 +114,34 @@
...
+}
+
+.use-parent-width {
+  max-width: none;
```
PeterBreen, line 142: This one is applying inconsistently. It works for a project page when that page is below the one-column breakpoint (for us, lg in bootstrap terms) but above that it leads to part of the page expanding and part not:
image

max-width should return to its prior behavior (controlled by the .container class) at that point -- try wrapping this rule in `@include media-breakpoint-down(lg) { ... }` and see if it works better? That should make it apply at lower breakpoints and disappear at higher ones, which means the page retains the designed widths correctly. The page layout/content has an upper bound for usability reasons and should be retained even in a framed view.

tools/iframe-test-resizer.html
```diff
@@ -0,0 +1,36 @@
+<!-- this file can be loaded directly from a browser and used to test embedding groups and projects within iframes
+see https://github.com/DemocracyLab/CivicTechExchange/issues/978 
+change groups/inframe to projects/inframe below to test embedding a project
+This if for testing with iframe resizer
+ -->
+<!DOCTYPE html>
+<html lang="en" style="height: 100%">
+  <head>
+    <meta charset="utf-8" />
```
PeterBreen, line 9: Missing the viewport meta tag here - makes testing go wrong without it.

tools/iframe-test.html
```diff
@@ -0,0 +1,32 @@
+<!-- this file can be loaded directly from a browser and used to test embedding groups and projects within iframes
+see https://github.com/DemocracyLab/CivicTechExchange/issues/978 
+change groups/inframe to projects/inframe below to test embedding a project
+This if for testing without using iframe resizer
+ -->
+<!DOCTYPE html>
+<html lang="en" style="height: 100%">
+  <head>
+    <meta charset="utf-8" />
```
PeterBreen, line 9: As above, need the viewport meta tag.

tools/iframe-test-resizer.html
```diff
@@ -0,0 +1,36 @@
...
+    <title>iframe test</title>
+  </head>
+  <body>
+    <div style="width: 100vw;">
+        <h1>This is a landing page with Group projects:</h1>
+        <style>
+          iframe {
+            width: 1px;
+            min-width: 100%;
```
PeterBreen, line 18: 100vw on the parent and 100% min-width here breaks mousewheel scroll - it forces the mouse into the iframe with no horizontal room to escape because the frame is filling all available width. Changing this to  `min-width: calc(100% - 30px);` so there's room for the mouse to move outside the frame would probably work. Removing the 100vw off the parent div may also be enough.

tools/iframe-test.html
```diff
@@ -0,0 +1,32 @@
...
+    <title>iframe test</title>
+  </head>
+  <body>
+    <div style="width: 100vw;">
```
PeterBreen, line 13: Same width concern as above. My thinking here is, we should have these test frames be as close to drop-in examples as possible, so a little polish that may not be required for internal dev isn't going to be wasted.

PeterBreen (changes requested):
I've made some comments on testing issues I've found, mostly on very small or very large devices. It's not exhaustive so please be sure to take a look at the whole layout on any test deployments (e.g. open san diego) on the whole range of small to very large viewports -- I'm sure there's going to be issues I didn't encounter using the test html files provided.

Broadly, dlab's site when displayed natively supports a mobile minimum of 360px device width up to a maximum viewport of 1440px (at which point the content is 1392px width and does not expand further) -- so please use those as your test boundaries for displaying dlab in a frame.

civictechprojects/static/css/partials/_Profile.scss
ddfridley: Okay, I found the Google Pixel Fold with a width of 280px that broke with 26rem.  As you suggested, I applied it in a breakpoint above small
image

As for the high res case, the parent web site will have their own breakpoints and we should avoid all the interactions.  In the open san diego case their content is contained in 960px container.  And I've seen similar containers on other brigade sites.   I suggest we avoid spending too much time perfecting how we render when it's really wide for an outlier use case, and focus on getting this feature out so that Mark can start pitching it to orgs.  After that, I expect we will get some real end-user feedback.   -- but I did fix the project display in the wide case - see that comment.

civictechprojects/static/css/partials/_Profile.scss
ddfridley: So I think I went the other way from what you were thinking.  I'm letting the project fill the full with of the viewport.  Based on the logic that the parent page will be imposing it's own container width, and breakpoints.

The fix was just to let "Profile-primary-container" go to width:100% if in an iframe. 

image

tools/iframe-test-resizer.html
ddfridley: Done

tools/iframe-test-resizer.html
ddfridley: I made the width change to the parent div. The min-width 100% is a recommendation from the iframe-resizer.

tools/iframe-test.html
ddfridley: done

tools/iframe-test.html
ddfridley: Happy to make the test-frames good examples, but this is the doc we are going to make available to people who want to try it: https://docs.google.com/document/d/1Uq5vLXrNzIEaXkpEw3X7K5DHvzETnPXKevCYG-O85KM/edit#heading=h.28dw2j9tlcg9

ddfridley (commented):
I have 3 pending comments so I am doing this to submit them.  
With this new push, I think this is ready to go.

civictechprojects/static/css/partials/_Profile.scss
PeterBreen: I tested that too, but no, text that wide is basically unreadable. Optimal line length for reading comprehension -- look at any newspaper, they do columns of text and not one fullpage line. Or most news sites constrain article width, etc. 

More generally, this page was designed deliberately, by the dlab design team, and so I think the default should be not to change that. Please revert this change so it has the same maximum content width as if it were displayed natively.

tools/iframe-test-resizer.html
PeterBreen: Sounds good, thanks. It was the combination of both -- so changing either is fine. If the resizer suggests min-width: 100% on the iframe, then let's keep that by all means.

civictechprojects/static/css/partials/_Profile.scss
PeterBreen: If you're interested, https://baymard.com/blog/line-length-readability is a good place to start.

ddfridley: The problem is that with the max-width there, even within a normal width viewport, we get a "AboutProjectDisplay" with blank-space margins on the sides. It's because these max-width breakpoints were set for the project withing the rest of the DLab html page, but now we are trying to provide the project as a component to be embedded in another website.

image

The site where this is being embedded won't want that much space on the sides and will want to control the padding/margin around our component themselves. 

We need to push the AboutProjectDisplay all the way to the edges of the viewport. 

I fully agree that the 3000px wide viewport is not useable - but the most common use case is that the developer using our iframe will put it under their own container div that will have it's own max-width.  I'm not overriding the max-width because I want it to go crazy wide, it's just about the extra margin on the sides. 

Here is the opensandiego example:
https://github.com/opensandiego/opensandiego.github.io/blob/master/_layouts/home.html#L32-L61

See that they already have their own container class https://github.com/opensandiego/opensandiego.github.io/blob/master/css/main.scss#L70

We can put something in the instructions telling developers that this should be used within a container no wider than 1392px, though I think it looks fine when a little wider too.

Or I can add a max-width if we are in an iframe to the MainController if we want to make it an enforced requirement.

What should we do?

PeterBreen: Hm, okay, I see what you're getting at. I agree, we don't want that whitespace there for narrower iframes. I apologize if I suggested that, I don't want it either! Only at very wide iframe width do I want to see changes. Thanks for explaining - I think we're agreeing on the optimal outcome here, it's just an implementation question?

Specifically,
- iframed profile page content is 100% of iframe width up to a defined breakpoint
- at that breakpoint, profile page content stops growing in width (so it'll have left/right whitespace)
  - (optionally) we can tell developers the iframe should probably have a max-width declared


If that's correct ... Well, first, after reading your comments, I don't think this is a blocking issue for release. Marlon - good to go on this. As you say it's very likely to be an edge case due to containers on the embedding website.

But on that note, I'd like to solve this on dlab's side. Then users don't have to do anything special, they just embed the iframe and it "just works."

Do you think it's a problem if some hypothetical user wants to render the world's longest row of project cards at 3000px width or whatever? If you do, maybe MainController is the place to put that max-width; I'd defer to you in that case. But if you don't think cards having unbound width is an issue, I'd be happy to try and figure out a targeted solution for the profile page only.  Let me know what approach you'd prefer - I'd be happy with either.

ddfridley: In a target solution for profiles, I figured out that I can just set max-width (in .use-parent-width) to 1392px instead of none.  And I will also change the name to .override-breakpoint-max-width.  So I don't need to do it in MainController.

But I will wait, unless you want this now.

PeterBreen: Let's wait; release testing's already started. I'm also thinking we may be able to approach this with Bootstrap's `.container-fluid` class, but that's definitely a wait-and-try-later level change.

PeterBreen (approved):

PeterBreen:
Approved all the parts of my review. Anything from your review, Marlon, I left alone.

ddfridley:
@marlonkeating I should have thought of this earlier.  I create a test server on heroku to use the iframe feature.
Please add
"https://dlab-iframe-test-0ac0b57c4ba3.herokuapp.com/" to the end of CSP_FRAME_ANCESTORS
and 
"*.dlab-iframe-test-0ac0b57c4ba3.herokuapp.com/" to the end of CSP_FRAME_SRC
in the production-mirror environment. 
Note that both those variables are arrays and it needs to fit the syntax.
[I don't know why I'm getting uuids on the urls now, but I guess they want to make it harder for the ECO servers"]
Should we add this to the production environment as well? 
And remember in production we need to add opensandiego.org

This will test the feature: https://dlab-iframe-test-0ac0b57c4ba3.herokuapp.com/

ddfridley:
update to above:
I pushed this branch to democracy-lab-dev and did the environment changes and testing there. 
---

common/components/utils/url.js
```diff
@@ -169,7 +169,7 @@ class urlHelper {
       return _url;
     } else {
       let _args: Dictionary<string> = {
-        prev: sectionArgs.section,
+        prev: sectionArgs.section!==Section.IframeProject ? sectionArgs.section : Section.AboutProject, // if login from within an iframe, go direct without iframe
```
marlonkeating, line 172: Do we want to have the same logic for groups?

Also, be sure to add new `logInThenReturn` unit tests to https://github.com/DemocracyLab/CivicTechExchange/blob/3f6985642c0992789d331eefc899ae8231b562ea/common/components/test/utils.test.js#L157

marlonkeating (approved):

PR #986, opened 2023-06-28
removes psycopg2 from requirements.txt to fix pip install failure

benj1r:


benj1r:
wrong branch to merge from, see #987.

PR #1001, opened 2023-07-09
change /admin to /secretadmin

chunlishao:
The new admin url:
image

The original admin url:
image

### DIFF
--- a/civictechprojects/urls.py
+++ b/civictechprojects/urls.py
@@ -77,7 +77,7 @@ urlpatterns = [
     re_path(r'^api/groups', views.group_search),
     re_path(r'^api/tags/groups', views.group_tags_counts),
     re_path(r'^api/tags', views.tags),
-    re_path(r'^admin/', admin.site.urls),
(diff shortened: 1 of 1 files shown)


civictechprojects/urls.py
```diff
@@ -77,7 +77,7 @@
     re_path(r'^api/groups', views.group_search),
     re_path(r'^api/tags/groups', views.group_tags_counts),
     re_path(r'^api/tags', views.tags),
-    re_path(r'^admin/', admin.site.urls),
+    re_path(r'^secretadmin/', admin.site.urls),
```
marlonkeating, line 80: If we're going to do this, the path should be configurable in a non-public place, such as an environment variable.  As it is, any attackers can simply look at the source code and see exactly where they need to go.

marlonkeating (changes requested):

civictechprojects/urls.py
chunlishao: good point. Will work on this.

chunlishao:
submitted a new pull request, closed this one.

PR #1007, opened 2023-07-17
change admin url to private url

chunlishao:
migrate the default admin url to a private url which configurated in environment variable for security.


chunlishao:
I put the url configuration in the example.env and environmentvariable.sh directly since it is used to load test data when developer set up the enrionment.

chunlishao:
will update the contribution doc when deployed.

democracylab/settings.py
```diff
@@ -551,3 +551,5 @@ def read_connection_config(config):
 MALICIOUS_FWD_PATTERNS = os.environ.get("MALICIOUS_FWD_PATTERNS", None)
 
 PASSWORD_RESET_TIMEOUT_DAYS = os.environ.get("PASSWORD_RESET_TIMEOUT_DAYS", 7)
+
+SECRET_ADMIN = os.environ.get("SECRET_ADMIN", "")
```
marlonkeating, line 555: Let's default to normal \admin path, so the site won't break if a developer forgets to set this environment variable.
```suggestion
SECRET_ADMIN = os.environ.get("SECRET_ADMIN", "admin")
```

marlonkeating (changes requested):

PR #1010, opened 2023-07-26
Update Home page to remove community partners and sustaining sponsors

marlonkeating:
On the home page, we're updating our sponsors listing by removing the Community Partners section, and removing the Sustaining sponsors from before.


PeterBreen (changes requested):
This looks good, but there's a few CSS cleanup items. Anything with the partner section CSS classes should should probably get removed as well. 
 - LandingController-partnersinaction
 - LandingController-partnersinaction-logo
 - LandingController-partnersinaction-container

I think those are the only three classes, and they're in /partials/_LandingController.scss. I don't see them used elsewhere on the site.

PeterBreen (approved):

Issue #1009, opened 2023-07-25
Throttle API calls to mitigate bot attacks


marlonkeating:
Lately we have been having issues with bot traffic, to the extent it has caused site outages recently. Similar to how we throttle front-end requests, we would like to throttle back-end api requests that are flooding in at greater rates than we would expect from an organic user.

## Requirements
- Requests are limited to 10/sec by default
- Rate limit can be configured via environment variable in order to clamp down during attack periods, or to relax/disable throttling for the purposes of performance tests.

## Resources
[Example PR: Add request throttling for anonymous and authenticated users](https://github.com/DemocracyLab/CivicTechExchange/pull/724)
https://www.django-rest-framework.org/api-guide/throttling/

PR #1012, opened 2023-08-02
Added api_view and throttle_class decorators to all views to throttle…

kennzi:
… back-end api calls


kennzi:
Closes #1009

civictechprojects/views.py
```diff
@@ -79,6 +79,8 @@ def group_tags_counts(request):
 
 # TODO: Pass csrf token in ajax call so we can check for it
 @csrf_exempt
+@api_view()
```
marlonkeating, line 82: Do we need the `@api_view()` lines?  If so, then POST methods like group_create need to be called with 'POST', because otherwise it defaults to just allowing GET.
```suggestion
@api_view(['POST'])
```

marlonkeating (changes requested):

civictechprojects/views.py
kennzi: From what I could find online, the @throttle_classes() decorator is meant to be used in conjunction with the @api_view() decorator and may not function correctly without it. I reviewed each of the views and updated the methods in the api_view() decorators, hopefully appropriately!

civictechprojects/views.py
```diff
@@ -139,6 +147,8 @@ def get_group(request, group_id):
         return HttpResponse(status=404)
 
 
+@api_view('GET', 'POST')
```
marlonkeating, line 150:
```suggestion
@api_view(['GET', 'POST'])
```
Page is failing to load with error message:
```
File "/app/civictechprojects/views.py", line 150, in <module>
@api_view('GET', 'POST')
TypeError: api_view() takes from 0 to 1 positional arguments but 2 were given
```

marlonkeating (changes requested):

civictechprojects/views.py
```diff
@@ -110,6 +114,8 @@ def group_edit(request, group_id):
 
 # TODO: Pass csrf token in ajax call so we can check for it
 @csrf_exempt
+@api_view(['GET', 'DELETE'])
```
marlonkeating, line 117: We actually use POST here, when we should have used DELETE, same with the other deletion operations.
```suggestion
@api_view(['GET', 'POST'])
```

marlonkeating (changes requested):

civictechprojects/views.py
```diff
@@ -110,6 +114,8 @@ def group_edit(request, group_id):
 
 # TODO: Pass csrf token in ajax call so we can check for it
 @csrf_exempt
+@api_view(['GET', 'POST'])
+@throttle_classes([AnonRateThrottle, UserRateThrottle])
 def group_delete(request, group_id):
```
marlonkeating, line 119: This operation is failing, and I'm not sure why.  It's definitely being sent with POST.

civictechprojects/views.py
```diff
@@ -366,6 +400,8 @@ def project_edit(request, project_id):
 
 # TODO: Pass csrf token in ajax call so we can check for it
 @csrf_exempt
+@api_view(['GET', 'POST'])
+@throttle_classes([AnonRateThrottle, UserRateThrottle])
 def project_delete(request, project_id):
```
marlonkeating, line 405: This operation also failing for unknown reasons.

marlonkeating (changes requested):

civictechprojects/views.py
agoncharov-reef: Can you elaborate what you mean by "failing"? :) I just ran it and it seems to work:
```
xh POST http://127.0.0.1:8000/api/groups/delete/2/
HTTP/1.1 204 No Content
```

c0ntribut0r:
@kennzi I think adding `@api_view` and setting `DEFAULT_THROTTLE_CLASSES` is a great solution. One thing to notice is that you don't need `@throttle_classes([AnonRateThrottle, UserRateThrottle])` because those throttle classes will be applied to any `ApiView` anyway, so I removed the decorator.

I added tests to some django views to ensure that throttling is working as expected. 

However, now that every view becomes `ApiView` (after adding an `@api_view` decorator), internal CSRF verification is done by `SessionAuthentication`, and `@csrf_exempt` does not work anymore. We could fix this by either patching `SessionAuthentication` or by fixing csrf. Since disabling csrf is not good for security anyway, I decided to patch js code to send csrf token. I ended up updating `apiHelper` and `ProjectAPIUtils` to send `X-CSRFToken` on `POST` requests, and it seems to work.

Also important - `def qiqo_webhook(request):` seems to be a webhook endpoint which 1) has no reason to be throttled, and 2) must use `@csrf_exempt`, otherwise `POST` request from 3rd-party service will be rejected. So I removed `@api_view` from it.

One thing left is modifying another js function to send CSRF token in headers - `export function deleteFromS3(s3Key)`. It uses raw `xhr` and tbh I don't know javascript at all to make it read the cookie value and send it as request header. Needs to be fixed by someone :(

marlonkeating (approved):
Looks great!  Props on fixing our long-standing CSRF issues 🥇

Issue #495, opened 2020-11-30
Render all page headers on backend, remove react-helmet component


glenbraun:
We recently added support for rendering meta headers on the backend, and should do that for all pages

**Checklist**

- [ ] Add page handlers in context_preload.py

- [ ] Remove Header component on frontend

- [ ] Remove React Helmet dependency

[Trello Card](https://trello.com/c/rg2bU5Gh)

PeterBreen:
Here's some code examples of how to do this, as several pages already have backend meta tag rendering:

For the About page, the tags are declared here
https://github.com/DemocracyLab/CivicTechExchange/blob/master/civictechprojects/helpers/context_preload.py#L87-L91

and then we use FrontEndSection to make sure that information goes where it belongs:
https://github.com/DemocracyLab/CivicTechExchange/blob/master/civictechprojects/helpers/context_preload.py#L183

And the about page has the correct meta information:
![image](https://github.com/DemocracyLab/CivicTechExchange/assets/16870466/12d1950a-96c2-4924-98e2-5b0458671307)

The component we want to remove, Headers.jsx (which uses React Helmet) is here: https://github.com/DemocracyLab/CivicTechExchange/blob/master/common/components/common/Headers.jsx and only used in two places. Unfortunately, they're the two most complicated to replace - they need to respond to whatever project or group data is loaded.

marlonkeating:
Fixed by https://github.com/DemocracyLab/CivicTechExchange/pull/1018

PR #1018, opened 2023-08-16
solve #495 render all page headers on backend remove react helmet component

QuanGiap:


civictechprojects/helpers/context_preload.py
```diff
@@ -144,6 +172,29 @@ def terms_preload(context, request):
...
+    context = default_preload(context, request)
+    context['title'] = 'Reset Password | DemocracyLab'
+    context['description'] = 'Reset password page'
+    return context
+
+def press_preload(context, request):
+    context = default_preload(context, request)
+    context['title'] = 'News Feed | DemocracyLab'
+    context['description'] =  "Read external articles and blog posts describing DemocracyLab's work."
+    return context
```
PeterBreen, line 197: The press page has been deleted, is there a reason we need this? I think it can be removed unless there's a technical reason to keep it.

civictechprojects/helpers/context_preload.py
```diff
@@ -191,7 +244,14 @@ def default_preload(context, request):
...
     {'section': FrontEndSection.Privacy.value, 'handler': privacy_preload},
-    {'section': FrontEndSection.Terms.value, 'handler': terms_preload}
+    {'section': FrontEndSection.Terms.value, 'handler': terms_preload},
+    {'section': FrontEndSection.SignUp.value, 'handler': sign_up_preload},
+    {'section': FrontEndSection.ContactUs.value,'handler':contact_us_preload},
+    {'section': FrontEndSection.ResetPassword.value,'handler':reset_password_preload},
+    {'section': FrontEndSection.FindEvents.value,'handler':find_events_preload},
+    {'section': FrontEndSection.FindGroups.value,'handler':find_groups_preload},
+    {'section': FrontEndSection.FindProjects.value,'handler':find_projects_preload},
+    {'section': FrontEndSection.Press.value,'handler':press_preload},
```
PeterBreen, line 254: As with the above comment about press.

common/components/controllers/LogInController.jsx
```diff
@@ -116,24 +116,28 @@ class LogInController extends React.Component<Props, State> {
...
+                </Button>
+              </a>
 
-              <Button
-                variant="link"
-                className="LogInController-createAccount"
-                onClick={url.navigateToSection.bind(this, Section.SignUp)}
+              {/* Need to be converted to link to be able to change the header metadata */}
+              <a
+                type="button"
```
PeterBreen, line 131: Do we need `type="button"` on this anchor element? As I understand it, anchor type accepts MIME types, not input types, so we don't. But please double check me.

civictechprojects/helpers/context_preload.py
```diff
@@ -111,6 +111,17 @@ def create_event_preload(context, request):
     context['description'] = 'Create event page'
     return context
 
+def create_group_preload(context, request):
+    context = default_preload(context, request)
+    context['title'] = 'Create an group | DemocracyLab'
```
PeterBreen, line 116:
```suggestion
    context['title'] = 'Create a Group | DemocracyLab'
```

this is an old mistake, not yours, but please fix this inconsistency with casing and grammar between group and project.

PeterBreen (changes requested):
I tested this for a little bit in my docker setup this morning - it looks pretty good, thank you! I've got a few line comments where there's some code we might not need and one grammar cleanup item. Otherwise, it seems to work just fine. I checked the changed pages and they're showing the correct page header info.

civictechprojects/helpers/context_preload.py
QuanGiap: > The press page has been deleted, is there a reason we need this? I think it can be removed unless there's a technical reason to keep it.

Hello Peter, thanks for reviewing the code and giving suggestions. It should be fine to delete the press page, I didn't know that the press page was deleted

QuanGiap:
> I tested this for a little bit in my docker setup this morning - it looks pretty good, thank you! I've got a few line comments where there's some code we might not need and one grammar cleanup item. Otherwise, it seems to work just fine. I checked the changed pages and they're showing the correct page header info.

Hello Peter, thanks for reviewing the code and giving suggestions. It should be fine to delete the press page, I didn't know that the press page was deleted

PeterBreen (approved):

marlonkeating (approved):

PR #1019, opened 2023-08-19
Fix #477 : enable hot reloading

### LINKED ISSUE #477: Enable Hot Reloading
Currently we need to run npm run build/dev in order to build front-end changes. If builds happened automatically when changes are made, it would streamline the development process, especially for development environments like Docker where running the build command is more laborious.

https://trello.com/c/G3LA9fka

AM-77:
fix #477

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -50,9 +50,7 @@ services:
     build: .
     env_file:
       - ./.env
-    command: bash -c '. /.nvm/nvm.sh use default &&
-                      npm run dev &&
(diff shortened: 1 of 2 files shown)


Yyao5:
Hello @AM-77, thank you for the PR! I tried this with my local docker env on Mac but the page did not seem to auto-reload/recompile. I have also tried adding watch configurations but didn't work. The PR did get rid of the error when running `npm run watch`.  Would you mind explaining why adding another frontend service and/or connecting over Zoom briefly?

Issue #1016, opened 2023-08-14
Authenticated user can go to the pages that are for Non Authenticated user


QuanGiap:
Authenticated users can go to the login page or any page that is for Non-Authenticated users. We can check by typing the whole URL while logged in, for example, "https://democracylab.org/login" 
![image](https://github.com/DemocracyLab/CivicTechExchange/assets/81385914/95331360-3d92-4c3e-a0ea-b7ab9652fb34)

The user should be redirected to the Home page if they go to those URLs

PR #1020, opened 2023-08-20
Fix issue #1016 Authenticated users can go to the pages that are for Non Authenticated user

QuanGiap:


AM-77:
@QuanGiap thank you for the PR.

just a suggestion, I think the url access restriction to /login and /signup when the user is already logged in should happen on the backend instead of the frontend, that way we dont even allow the user to load the page at all.

QuanGiap:
> @QuanGiap thank you for the PR.
> 
> just a suggestion, I think the access restriction to /login and /signup when the user is already logged in should happen on the backend instead of the frontend, that way we dont even allow the user to load the page at all.

QuanGiap:
> @QuanGiap thank you for the PR.
> 
> just a suggestion, I think the access restriction to /login and /signup when the user is already logged in should happen on the backend instead of the frontend, that way we dont even allow the user to load the page at all.

That is a good suggestion, but it can bypass the check if the front end uses urlHelper.navigateToSection(). It uses history.pushState() which causes the browser won't attempt to load the given URL.

common/components/controllers/LogInController.jsx
```diff
@@ -41,6 +41,13 @@ class LogInController extends React.Component<Props, State> {
     }
   }
 
+  //if the user is logged in, navigate back to home page
+  componentDidMount():void{
+    if (CurrentUser.isLoggedIn()) {
+      document.location.href = url.section(Section.Home);
+    }
```
marlonkeating, line 48: Let's make this its own function in [url.js](https://github.com/DemocracyLab/CivicTechExchange/blob/2a53c9645944133f6964cde357eed18d309b1f74/common/components/utils/url.js#L70)

marlonkeating (changes requested):

marlonkeating (changes requested):
Refactor looks good, but would be good to add a unit test for the helper method at https://github.com/DemocracyLab/CivicTechExchange/blob/024bd76b3b526b2f94c620f08315caff3081c5af/common/components/test/utils.test.js#L98

PR #1022, opened 2023-08-21
995 Fix Login Button Redirect Issue

Tong-Ding:
### Problem
The issue arises when users log in from http://0.0.0.0:8000/login, leading them to http://0.0.0.0:8000/. If users navigate to other pages and then click the login button, a prev parameter may be included. However, when logging in from URLs containing the prev parameter, such as http://0.0.0.0:8000/login?prev=Home, they are redirected to http://127.0.0.1:8000/, necessitating **a second login**. This inconsistent behavior creates a confusing user experience.

The underlying problem stems from the development server starting at http://0.0.0.0:8000/, while the PROTOCOL_DOMAIN set in the Dockerfile is http://127.0.0.1:8000. Changing the starting server to 127.0.0.1 is not an option, as it would prevent connection to the server via a browser.

### Solution
The optimal solution to this issue involves the creation of a Middleware that will automatically redirect requests starting with 0.0.0.0 to 127.0.0.1. Essentially, if users access http://0.0.0.0:8000/groups, they will be redirected to http://127.0.0.1:8000/groups; similarly, http://0.0.0.0:8000/login will redirect to http://127.0.0.1:8000/login. This approach maintains the integrity of the code while delivering a redirection process that is more efficient than handling it within the view function.

### DIFF
--- a/common/helpers/redirectors.py
+++ b/common/helpers/redirectors.py
@@ -1,6 +1,11 @@
 from django.shortcuts import redirect
 from urllib.parse import urlparse
-from common.helpers.front_end import clean_invalid_args, get_clean_url, get_page_section, redirect_from_deprecated_url
+from common.helpers.front_end import (
+    clean_invalid_args,
+    get_clean_url,
+    get_page_section,
+    redirect_from_deprecated_url,
+)
 
 
 class RedirectTo(Exception):
@@ -13,7 +18,10 @@ class RedirectMiddleware:
         self.get_response = get_response
 
     def __call__(self, request):
-        # Nothing to do here, just using this middleware for redirect exception handling
+        # Redirect to localhost if the host starts with '0.0.0.0'
+        if request.get_host().startswith("0.0.0.0"):
+            new_url = request.build_absolute_uri().replace("0.0.0.0", "127.0.0.1", 1)
+            return redirect(new_url)
         return self.get_response(request)
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


Tong-Ding:
> ### Problem
> The issue arises when users log in from http://0.0.0.0:8000/login, leading them to http://0.0.0.0:8000/. If users navigate to other pages and then click the login button, a prev parameter may be included. However, when logging in from URLs containing the prev parameter, such as http://0.0.0.0:8000/login?prev=Home, they are redirected to http://127.0.0.1:8000/, necessitating **a second login**. This inconsistent behavior creates a confusing user experience.
> 
> The underlying problem stems from the development server starting at http://0.0.0.0:8000/, while the PROTOCOL_DOMAIN set in the Dockerfile is http://127.0.0.1:8000. Changing the starting server to 127.0.0.1 is not an option, as it would prevent connection to the server via a browser.
> 
> ### Solution
> The optimal solution to this issue involves the creation of a Middleware that will automatically redirect requests starting with 0.0.0.0 to 127.0.0.1. Essentially, if users access http://0.0.0.0:8000/groups, they will be redirected to http://127.0.0.1:8000/groups; similarly, http://0.0.0.0:8000/login will redirect to http://127.0.0.1:8000/login. This approach maintains the integrity of the code while delivering a redirection process that is more efficient than handling it within the view function.

Issue #1014, opened 2023-08-06
Restrict volunteering with project to only listed roles


marlonkeating:
Due to user feedback, we would like to restrict users from volunteering for projects with roles that aren't currently listed on the project.

## Background discussions
https://github.com/DemocracyLab/CivicTechExchange/issues/938

## Acceptance Criteria
- Do not show 'Volunteer with Project' button on project profile if the project has no open roles
- For a project with roles, upon clicking 'Volunteer with Project', do not list the 'Other' option (which currently lets them select any role)

## Code Locations
[Volunteer with Project Button](https://github.com/DemocracyLab/CivicTechExchange/blob/2667ae0c3dfc9f344a4d009f1cc894930e5417f9/common/components/common/projects/ProjectVolunteerButton.jsx#L153)

['Other' role option to remove](https://github.com/DemocracyLab/CivicTechExchange/blob/2667ae0c3dfc9f344a4d009f1cc894930e5417f9/common/components/common/projects/ProjectVolunteerModal.jsx#L47)

qchiujunhao:
Completed by PR #1023

PR #1023, opened 2023-08-21
#1014 Restrict volunteering with project to only listed roles

qchiujunhao:
- [x] For a project with roles, upon clicking 'Volunteer with Project', do not list the 'Other' option (which currently lets them select any role)
- [x] Do not show 'Volunteer with Project' button on project profile if the project has no open roles

### DIFF
--- a/common/components/common/projects/ProjectVolunteerButton.jsx
+++ b/common/components/common/projects/ProjectVolunteerButton.jsx
@@ -151,28 +151,34 @@ class ProjectVolunteerButton extends React.PureComponent<Props, State> {
   }
 
   _renderVolunteerButton(): React$Node {
-    return this.state.isAlreadyVolunteering ? (
+    if (this.state.isAlreadyVolunteering) {
       // TODO: Make this its own component and hook up to My Projects page
-      <Button
+      return (<Button
         className="AboutProject-button"
         type="button"
         variant="destructive"
         onClick={this.handleShowLeaveModal}
       >
         Leave Project
-      </Button>
-    ) : (
-      <Button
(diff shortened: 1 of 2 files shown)


common/components/common/projects/ProjectVolunteerButton.jsx
```diff
@@ -151,28 +152,36 @@ class ProjectVolunteerButton extends React.PureComponent<Props, State> {
...
-        title={this.state.buttonTitle}
-        onClick={this.handleShowJoinModal}
-      >
-        Volunteer With Project
-      </Button>
-    );
+      </Button>);
+
+    } else {
+        if (!_.isEmpty(this.state.project.project_positions)) {
```
marlonkeating, line 167: Didn't call this out in the acceptance criteria, but we would also like to hide the button in the case that all positions are hidden (`project_position.isHidden === true`)

marlonkeating (changes requested):

common/components/common/projects/ProjectVolunteerButton.jsx
qchiujunhao: > Didn't call this out in the acceptance criteria, but we would also like to hide the button in the case that all positions are hidden (`project_position.isHidden === true`)

will do

marlonkeating, line 167:
```suggestion
        if (!_.isEmpty(this.state.project.project_positions) && this.state.project.project_positions.some(position => !position.isHidden)) {
```

marlonkeating (approved):

PR #1029, opened 2023-08-30
fix: compress frontend files locally with webpack

### LINKED ISSUE #968: Webpack build/serve compressed bundle files
We use webpack to create three js bundle files on build:
 - runtime.bundle.js
 - main.bundle.js
 - vendors.bundle.js 

See PR #346 for context.

These are served via straight hyperlinks on the one html file which django serves and react uses as a base for everything else:
https://github.com/DemocracyLab/CivicTechExchange/blob/7d28862876c0142e0dc0f8957fd44b76b3a1b7eb/civictechprojects/templates/new_index.html#L16-L18

However, we are not using gzip or brotli or any compression. According to heroku, [that's our app's responsibility](https://devcenter.heroku.com/articles/compressing-http-messages-with-gzip) and we're getting rightly knocked in Lighthouse reporting for not doing this. 
image

Therefore: in order to reduce initial load time and transfer, we should offer compressed bundles. Gzip, at minimum. Webpack should be able to build them as part of `npm run build` and then we need to offer the compressed version to any browser requesting it with the uncompressed bundles as fallback options. Right now the app doesn't care if an http request has `Accept-Encoding:` because we don't have any encoded bundles to offer and don't have a way to send them since the hrefs are just hrefs.

As you can see in the lighthouse report I posted our compiled sass styles.css file is also something we can offer this way and I'd like to see that too, but the bundles are a higher priority.

To sum up,
 - configure webpack to produce compressed bundles on build, as well as uncompressed
 - serve these if the http request has the appropriate accept-encoding header
 - serve uncompressed bundles as fallback if needed
 - don't break anything ;)

abnud1:
Fixes #968 

This PR builds two compressed versions of static files: a brotli version and a gzip version.

When testing the application through gunicorn, It's verified through browser's dev tools Network tab that the compressed version is served:
democracylab-brotli

And gzip version is served when `Accept-Encoding` header doesn't contain `br` which happens with chrome when server is not HTTPS:
democracylab-gzip

I installed `compression-webpack-plugin` v6 because the latest version isn't compatible with webpack v4 which is the version we're using now, in another PR I plan to solve this issue by upgrading webpack to v5 and its loaders and plugins.

The compressed files won't be served with `python manage.py runserver` since Django's dev server isn't made for that purpose, you have to run the application using `gunicorn`, also compression is only enabled in `webpack.prod.js` so `npm run dev` won't work, you have to use `npm run build`, I managed to accomplish all that by changing `docker-compose.yml` file at this line:
before
```yaml
   command: bash -c '. /.nvm/nvm.sh use default &&
                      npm run dev &&
                      python manage.py runserver 0.0.0.0:8000'
```
after
```yaml
   command: bash -c '. /.nvm/nvm.sh use default &&
                      npm run build &&
                      gunicorn -b 0.0.0.0:8000 democracylab.wsgi'
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,12 +28,19 @@ var/
 .installed.cfg
 *.egg
 *.js.map
+*.js.map.br
+*.js.map.gz
 *.css.map
+*.css.map.gz
+*.css.map.br
 main.styles.css
-
+main.styles.css.gz
+main.styles.css.br
 # Files that are now built automatically during heroku deployment
 bundle.js
 *.bundle.js
+*.bundle.js.gz
+*.bundle.js.br
 styles.css
 
 
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "aws-sdk": "^2.289.0",
     "babel-loader": "^8.3.0",
     "child_process": "^1.0.2",
+    "compression-webpack-plugin": "6",
     "core-js": "^3.1.4",
     "fs": "0.0.1-security",
     "indent-string": "2.1.0",
--- a/webpack.prod.js
+++ b/webpack.prod.js
@@ -2,6 +2,7 @@ const merge = require('webpack-merge');
 const common = require('./webpack.common.js');
 const TerserPlugin = require('terser-webpack-plugin');
 const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin');
+const CompressionPlugin = require("compression-webpack-plugin");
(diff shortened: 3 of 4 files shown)


Rafinad001 (approved):

PeterBreen:
Just so I'm clear - would this require changing how we serve files on production (needing gunicorn instead of django's runserver) as well, or is that a dev environment specific change?

PR #1037, opened 2023-09-05
Fix "cannot access body after reading from request's data stream"

c0ntribut0r:
This PR substitutes access to `request.body` with `request.data`, as per official documentation: https://www.django-rest-framework.org/api-guide/requests/#data


marlonkeating (approved):
Confirmed fixes work, thanks!

PR #1051, opened 2023-10-09
update 'Country' to 'Country/Region' in Location in "Find Projects"

hyihua:


PeterBreen:
Maybe I missed this in a team meeting, but why make this change?

PeterBreen:
Label changes are managed by a ux design team member; I'll run this by the design lead and let them follow up on it. If nothing else we have to keep our design files and deployed site in sync.

PeterBreen:
Our design lead got back to me and is fine with `Country / Region` so that change is good to go. I'd request a space between it and (Required), just in case we need to have a line break on a narrow device. `Country/Region (Required)` 

Thanks for spotting this.

hyihua:
Created an account and a project for replicating this issue. It turns out that the code snippet listed above cannot fix this issue.

hyihua:
I figured out the reason that why the code snippet I found didn't work for this issue. The reason is that the we only use the `is_searchable` attribute to filter projects but didn't make filtering for the country list, and that's why we can still see countries with no projects in the country selector.

CC: @PeterBreen  @marlonkeating

hyihua:
@PeterBreen I think this bug is fixed and ready for review now. I created a project for testing and it was pending for approval. But it was not shown in the country list.

image

image

PR #1075, opened 2023-11-20
Impact Dashboard - Cmu Practicum 23

SebZhuC:


civictechprojects/views.py
```diff
@@ -1278,53 +1583,171 @@ def get_testimonials(request, category=None):
...
+        "yearly_stats" : stats_list,
+        "cumulative_renewal_percentage": cumulative_renewal_percentage,
+        "volunteer_matching" : volunteer_matching
+    }
+
+    return JsonResponse(data)
+
+@api_view()
+def volunteer_roles(request):
+    volunteers = VolunteerRelation.objects.filter(is_approved=True)
```
marlonkeating, line 1716:
```suggestion
    volunteers = VolunteerRelation.objects.filter(is_approved=True, deleted=False)
```

common/components/controllers/AggregatedDashboardController.jsx
```diff
@@ -0,0 +1,95 @@
...
+    this.state = {
+        volunteerStats: null,
+        retryCount: 0,
+        maxRetries: 3,  // Maximum number of retries
+        retryDelay: 1000 // Delay between retries in milliseconds
+    };
+  }
+
+  componentDidMount(): void {
+    document.getElementById('detailButton').style.display = 'none';
```
marlonkeating, line 27: TODO: Inline style instead

example.env
marlonkeating, line 1: Why is this being deleted?

oauth2/apps.py
```diff
@@ -1,5 +1,4 @@
 from django.apps import AppConfig
 
-
```
marlonkeating, line 3: Extraneous space

marlonkeating (changes requested):

common/components/controllers/AggregatedDashboardController.jsx
marlonkeating, line 1: There are serious performance issues with this page as it stands.  Multiple components have their own separate API calls, which on aggregate result in a long, fragmented page load where individual charts load at different times.  For the sake of performance, I recommend putting all the data behind a single API call (whose payload should be cached since the data does not change often).

marlonkeating (changes requested):

common/components/controllers/AggregatedDashboardController.jsx
typesend: Resolved in #1088

typesend:
Closing this PR as it lives on as #1088

PR #1082, opened 2023-12-31
1079 filter unapproved groups

qchiujunhao:


qchiujunhao:
Can be fixed in #1051, so close this one now!

PR #1116, opened 2024-10-14
Fix: Ensure EditUserModal can't be closed until the crop is finished or canceled

### LINKED ISSUE #896: Perceived failure to save profile picture

Nostromos:
Fixes #896 and very similar to #911.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,5 @@
 default_language_version:
-    python: python3.8
+    python: python3.11
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.3.0
--- a/common/components/common/FileInfo.jsx
+++ b/common/components/common/FileInfo.jsx
@@ -1,3 +1,5 @@
+// @flow
+
 export type FileInfo = {|
   id: ?number,
   key: string,
--- a/common/components/common/upload/S3Data.jsx
+++ b/common/components/common/upload/S3Data.jsx
@@ -1,3 +1,5 @@
+// @flow
+
 export type S3Data = {|
   url: string,
   fields: {


ddfridley:
@Nostromos to fix the issue you are working on, is it necessary to make these changes?  Changing the version of python impacts the whole system, and would require more extensive QA testing before release.  As for the @flow changes, same question, but I don't know @flow and if effects run time, or just build time.  I know these seem like simple changes, but this is a product in production and we need to be very careful about changes.   QA testing and releasing to production is a big task, we want to make the changes as focused as possible. Thanks!

Nostromos:
Hey @ddfridley - Thanks for the engagement! FYI this is still a draft so not ready to review yet. Childcare fell through today so halfway through my commits had to stop and go home. 

> Changing the version of python impacts the whole system, and would require more extensive QA testing before release

This only changes the python version for the `pre-commit` package config, not python for the whole system. I could not get working otherwise and `pre-commit` is required to commit anything. I plan to remove this commit before marking this PR ready for review. I'll take some time after to figure out why `pre-commit` isn't recognizing my python 3.8 install. 

> As for the @flow changes, same question, but I don't know @flow and if effects run time, or just build time. I know these 
> seem like simple changes, but this is a product in production and we need to be very careful about changes. QA testing and releasing to production is a big task, we want to make the changes as focused as possible. 

We're aligned. These flow changes *shouldn't* affect run or build time as far as I can tell. Reviewing my own builds doesn't show any change.

---

I'm being really careful not to change more than necessary and to keep the fix in scope. I'm also referring to the other PR submitted for this issue and getting ahead of the issues raised there. Honestly, that fix looked good but I need the practice and its good exposure to the codebase for me. 

tl;dr Check back when this PR is out of draft.

Nostromos:
**Update:** We managed to find childcare so I should be back at this tomorrow. Major apologies for the delay but kiddo is really little so its nonstop.

ddfridley:
@Nostromos  Sorry for the slow responses. I'm back from vacation and so is Marlon.  How is it going?

PR #1171, opened 2026-06-17
Mailchimp safe 1169

ddfridley:
## Problem

We were having trouble with bot emails ending up in mailchimp, and we are closing vulnerability signup and newsletter opt-in flow.
- Users could be sent to Mailchimp before verifying their DemocracyLab email address.
- Signup lacked strong server-side bot protections.
- reCAPTCHA requests were blocked by CSP in dev, which could prevent successful signup.
- The footer newsletter subscribe button pointed to an invalid Mailchimp form URL.

In addtion:
- Docker builds failed on fresh machines because of an outdated Yarn apt signing key.

## What This PR Changes

- Defers Mailchimp subscription until after DemocracyLab email verification succeeds.
- Adds server-side signup abuse controls:
  - reCAPTCHA token verification
  - IP-based signup rate limiting
- Updates Mailchimp member status to subscribed after app-level verification to avoid a redundant second confirmation email.
- Replaces the broken footer newsletter endpoint with the current hosted Mailchimp signup URL.
- Adds required CSP `connect-src` entries for reCAPTCHA endpoints and documents why.
- Adds Mailchimp setup guidance in code to reduce future configuration drift.

## Dev Server Testing
On the dev server we had to set 

- MAILCHIMP_API_KEY - to a new key because the old one was shutdown. 
- MAILCHIMP_SUBSCRIBE_LIST_ID - need to make sure what's in production is this.
- FAKE_EMAILS to False to generate emails,
- REDIS_ENABLED to False because no worker is being started in dev to process the request queue.


ddfridley:
Claude generated the change to Dockerfile - and I had Gemini Pro review it for security since it's pretty opaque to me. But I could get docker to build on a fresh machine without this.

I have tested user signup on the dev server and it works and doesn't add users to mailchimp until after they verify their email.

ddfridley:
I added common/static to .gitignore because in my other branches webpack and storybook are generating lots of files there, but these files shouldn't be added to the repo.

democracylab/views.py
```diff
@@ -17,6 +20,55 @@
 from salesforce import contact as salesforce_contact
 
 
+def _client_ip(request):
+    forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR')
+    if forwarded_for:
+        return forwarded_for.split(',')[0].strip()
+    return request.META.get('REMOTE_ADDR', '')
```
marlonkeating, line 30: [`X-Forwarded-For` is a client-controlled header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For). Any attacker can set `X-Forwarded-For: 1.2.3.4` to cycle through arbitrary IPs and evade the rate limit entirely — this exact attack pattern is documented in [The Great Rate Limit Escape](https://medium.com/@InfoSecTube/the-great-rate-limit-escape-an-x-forwarded-for-exploit-story-709eb2331aae) and [Beware of X-Forwarded-For](https://www.stackhawk.com/blog/do-you-trust-your-x-forwarded-for-header/). It has also been assigned a CVE against at least one Python framework ([Litestar GHSA-hm36-ffrh-c77c](https://github.com/litestar-org/litestar/security/advisories/GHSA-hm36-ffrh-c77c)).

Per the [Heroku HTTP Routing docs](https://devcenter.heroku.com/articles/http-routing), Heroku's router *appends* the detected client IP to the right side of the existing list. This means a spoofed value supplied by the client appears at index `[0]` (left), while the router-added trusted value is at the *rightmost* position. The current code reads `[0]` — the attacker-controlled slot. See also [What is X-Forwarded-For and when can you trust it?](https://httptoolkit.com/blog/what-is-x-forwarded-for/) for a clear walkthrough of the left-vs-right trust model.

**Recommendation:** Trust only the *last* IP appended by the reverse proxy (the one Heroku adds), not the first. On Heroku, the real client IP is the *last* value in the `X-Forwarded-For` chain:

```python
def _client_ip(request):
    forwarded_for = request.META.get('HTTP_X_FORWARDED_FOR')
    if forwarded_for:
        return forwarded_for.split(',')[-1].strip()  # last = trusted proxy's append
    return request.META.get('REMOTE_ADDR', '')
```

democracylab/settings.py
marlonkeating, line 1: The rate limiter uses `django.core.cache`. The PR's own testing notes say `REDIS_ENABLED` was disabled during development. Without Redis, Django falls back to the `locmem` (local-memory) cache, which is a dict stored in a single Python process and [explicitly not suitable for rate limiting in multi-process deployments](https://django-ratelimit.readthedocs.io/). On Heroku with multiple dynos, each dyno holds an independent counter — a bot can exceed the limit N×dyno-count times by round-robining across dynos. This is a well-known failure mode: [Django Throttling with DRF and Redis](https://medium.com/@connect.hashblock/django-throttling-with-drf-and-redis-smart-rate-limiting-for-scalable-apis-3ee28c58fe2a) covers it in detail, and [django-defender issue #96](https://github.com/jazzband/django-defender/issues/96) shows the same bug in a comparable IP-blocking library.

**Recommendation:** Add a startup warning or assertion so the condition is visible in logs:

```python
# settings.py or apps.py ready()
import warnings
if not REDIS_ENABLED:
    warnings.warn("Rate limiting is non-functional without Redis (REDIS_ENABLED=False).")
```

Dockerfile
```diff
@@ -1,6 +1,13 @@
 # https://hub.docker.com/r/nikolaik/python-nodejs
 FROM nikolaik/python-nodejs:python3.10-nodejs16
 
+# Refresh Yarn repo GPG key (old key 62D54FD4003F6525 expired; fetch the
+# current one so apt-get update succeeds on bookworm-based images).
```
marlonkeating, line 5: I asked Claude about this change, and it came up with the following explanation.  If this is accurate, you might consider adding in some of these additional details to the comment.
```suggestion
# The nikolaik base image ships a Yarn apt source whose bundled GPG key
# (62D54FD4003F6525) expired, and Debian Bookworm dropped the legacy apt-key
# trust store in favour of per-source signed-by keyrings.  This block:
#   1. Downloads the current Yarn public key as a binary keyring.
#   2. Updates /etc/apt/sources.list.d/yarn.list to reference it via signed-by.
# Without this, `apt-get update` fails with "NO_PUBKEY 62D54FD4003F6525".
# To verify: docker build --no-cache and watch the apt-get update step.
```

democracylab/views.py
```diff
@@ -139,7 +196,17 @@ def verify_user(request, user_id, token):
         # TODO: Add feedback from the frontend to indicate success/failure
         contributor = Contributor.objects.get(id=user_id)
         contributor.email_verified = True
+        subscribe_to_newsletter = contributor.newsletter_signup_requested
+        contributor.newsletter_signup_requested = False
```
marlonkeating, line 207: nit: I see why we're setting this false when we're about to subscribe the user to the newsletter, but it took some reading.  Additional comments would help readers understand the logic flow.

common/helpers/mailing_list.py
```diff
@@ -1,3 +1,33 @@
+"""
+Mailchimp setup instructions
```
marlonkeating, line 2: Good documentation here!  Can we get it added to our Contributor Guide?

common/helpers/mailing_list.py
```diff
@@ -16,11 +46,19 @@ def print_error(self, err_msg):
         err_msg = "Failed to subscribe {first} {last}({email}) to mailing list: {err_msg}".format(
             first=self.first_name,
             last=self.last_name,
-            email=self.email,
+            email=self.masked_email(),
             err_msg=err_msg,
         )
         print(err_msg)
 
+    def masked_email(self):
```
marlonkeating, line 32: Can we get unit test for this email masking behavior?  Can put it in [test_helpers.py](https://github.com/DemocracyLab/CivicTechExchange/blob/master/common/tests/test_helpers.py)

marlonkeating (changes requested):
Reviewed with help of Claude.

One thing I found during testing on the dev environment, registering threw an error that looks related to database migration when I attempted to sign up:
`Jun 28 09:06:36 AM democracy-lab-dev app/web.1 django.db.utils.ProgrammingError: column "newsletter_signup_requested" of relation "democracylab_contributor" does not exist`

ddfridley:
@marlonkeating 
> Reviewed with help of Claude.
> 
> One thing I found during testing on the dev environment, registering threw an error that looks related to database migration when I attempted to sign up: `Jun 28 09:06:36 AM democracy-lab-dev app/web.1 django.db.utils.ProgrammingError: column "newsletter_signup_requested" of relation "democracylab_contributor" does not exist`

I had this problem too, until I ran `heroku run python manage.py migrate democracylab` but since I ran it, it shouldn't have to be run again.  Is there some other server accessing the same database? 

Claude is saying: "his error is happening because the database schema on democracy-lab-dev is out-of-sync and the migration wasn't physically applied to that database during the deploy. We can fix the dev environment by running heroku run python manage.py migrate democracylab or by temporarily reverting and re-applying the migration history on that shared DB."

I'm looking into it, but wondering if sharing the same DB with some other server?
Thanks,

common/helpers/mailing_list.py
ddfridley: see test_helpers.py

common/helpers/mailing_list.py
ddfridley: added, and this file now links to that file so we don't have redundant documentation to get out of sync.

democracylab/settings.py
ddfridley: added

democracylab/views.py
ddfridley: added

democracylab/views.py
ddfridley: explained.

Dockerfile
ddfridley: added

ddfridley:
The fix for the migration issues was to auto generate migrations for the change to Django 4.   The logs show the migrations working now.  See those files for a more detailed explanation.

I have pushed these changes dev and it's working.

marlonkeating (approved):

